PR #241, opened 2019-10-31
updating generate to comment out description in transforms

patel-bhavin:
btool says invalid key "description" , commenting that key as a fix


P4T12ICK:
Looks good.

PR #242, opened 2019-10-31
Code refactoring generate script

P4T12ICK:
Change Notes:
- Moved to jinja2 templating for generating Configuration files
- Refactored generate.py sourcecode
- Fixed some bugs
- Changed some wrong UUIDs, which didn’t match the name

Testing:
- tested the old generate.py against the new version and used diff to ensure that the results are the same.


patel-bhavin:
the python looks good! Let's fix the conflicts!

PR #257, opened 2019-11-14
Fixed package updates

trogdorsey:
requests was throwing an error, updated subpackages for it


patel-bhavin:
Dorsey tested it and it looks good

PR #278, opened 2019-12-11
CRL_1700 ES Macros

d1vious:
- [x] introduced `security_content_ctime` replacing the ctime macro across all content
- [x] introduced `security_content_summariesonly` replacing the summariesonly macro across all content
- [x] removed runstory macro definition
- [x] removed comment macro for  empty definitions
- [x] updated macro jinja template for output empty definitions
- [x] updated searches/files to use standard naming conventions vs camel case


patel-bhavin:
updated as per dorsey's feedback, fixed some other files too. updated searches for naming consistency as well, while i was at it

patel-bhavin:
@trogdorsey : Can you perhaps review this? want to make sure i didnt miss anything

P4T12ICK:
@patel-bhavin @trogdorsey : Should the macro security_content_ctime not be independent from ctime (not use ctime)? Otherwise, ES needs to be installed to run our searches. In my opinion we need to put the content of ctime in the macro and not a reference to ctime.

P4T12ICK:
@patel-bhavin I found an issue with the comment macro (which is used for empty pre and post filter). If you want to run the search in any app besides search, the searches will fail because the comment macro has sharing in app and is defined in search app. We should ship with a config which makes the comment macro global accessible.

d1vious:
> @patel-bhavin @trogdorsey : Should the macro security_content_ctime not be independent from ctime (not use ctime)? Otherwise, ES needs to be installed to run our searches. In my opinion we need to put the content of ctime in the macro and not a reference to ctime.

`ctime` is offered by core as a command: https://docs.splunk.com/Documentation/Splunk/latest/SearchReference/Convert#Convert_functions

patel-bhavin:
> @patel-bhavin I found an issue with the comment macro (which is used for empty pre and post filter). If you want to run the search in any app besides search, the searches will fail because the comment macro is shared only inside the search app. We should ship with a config which makes the comment macro global accessible.

@P4T12ICK : removed the comment macro all together after chatting with @d1vious

PR #289, opened 2019-12-17
CRL-1698 new version credential dumping

P4T12ICK:
New version of credential dumping analytics story including the following detections:
- Access LSASS Memory for Dump Creation
- Create Remote Thread into LSASS
- Detect Credential Dumping through LSASS access
- Unsigned Image Loaded by LSASS
- Attempted Credential Dump From Registry via Reg.exe
- Detect Mimikatz Using Loaded Images
- Attempt To Set Default PowerShell Execution Policy To Unrestricted or Bypass
- Creation of Shadow Copy with ntdsutil
- Creation of Shadow Copy with vssadmin
- Creation of Shadow Copy with wmic and powershell
- Creation of Shadow Copy with wmic
- Credential Dumping via Copy Command from Shadowcopy
- Credential Dumping via Symlink to Shadowcopy


P4T12ICK:
Thank you for your great feedback:
1) will have a look
2) in my opinion they were not really helpful. I tried to be more specific to the analytics story.
3) in my opinion, we can only add it if we have tested it.
4) will change that
5) Good idea. Will try to combine them.
6) thank you. I will rename the macro.
7) What is the intro macro?
8) I renamed a few because I was going crazy because I didn't find the searches when the name of the search and the name of the file was completely different. There should be no impact to the customer when the file name changed because it will still generate the same savedsearches.conf
9) good idea. Will test it.
10) yes she is doing it.

rvaldez617:
7 - the macro I'm referring to would be the appropriate version of `wineventlog_sysmon` for windows events
8 - cool, again, just don't think we've actually done this, so unsure of any potential impacts. Fine with a name tweak in general

P4T12ICK:
I worked on the feedback of Rico:
1) I totally agree to your feedback
2) I decided to develop more tailored investigations. I think we have to rethink the investigations once we know how they get executed in Mission Control.
3) I think that is also a topic, which should be discussed in the team. I'm not happy with releasing something which I was not able to test.
4) changed. Thanks for the feedback.
5) Combined three searches to one. Great idea.
6) renamed to `sysmon`
7) introduced macro `wineventlog_security` for the investigations
8) already discussed
9) changed and tested it. This is definitely the better default value.
10) Working with Brianna on it.

PR #291, opened 2019-12-18
removing index and keeping source

patel-bhavin:
removing index from the macro file and adding instructions in the descriptions for customizing
currently leveraging sourcetype and eventtype where applicable 

Note: This still would work OOTB in the range since we need "index=*" before thee search executes, opened an issue for that :https://github.com/splunk/attack_range/issues/151

still working on a good solution, but this should work for now


P4T12ICK:
I'm fine with that, but when a customer defines index=something. This will lead to the same problems with ASX, which should be fixed in ASX.

P4T12ICK:
Tested it with the attack range. Approved. Detections still working as expected.

PR #315, opened 2020-01-18
CRL-1716 - Fixing by clause to allow searching Processes.process for …

jcbrewer5D6400:
…obfuscation chars.


detections/malicious_powershell_process_obfuscation_techniques.yml
```diff
@@ -35,7 +35,7 @@ detect:
       search: '| tstats `security_content_summariesonly` count values(Processes.process) as process
```
patel-bhavin: add a post filter macro

patel-bhavin: add a macro file

jcbrewer5D6400: These are done.

PR #316, opened 2020-01-18
CRL-1717 - Added "search *" definitions to the macros without them… resolves #292

### LINKED ISSUE #292: Empty macros
We are shipping empty macros in our app and that breaks the detection. We need to figure out a better solution for shipping macros that we expect our users to customize

jcbrewer5D6400:
…ions.


patel-bhavin:
@jcbrewer5D6400 : gonna hold of on this PR. need to discuss with the team if investigation objects need macros

PR #318, opened 2020-01-21
CRL-1495 update share creation and deletion

jzsplunk:
@patel-bhavin helped review the logic here and suggested moving the hidden share specific filters to an optional step that reflects the more general case of detecting any type of share delete/create activity.  This PR has renamed the origional detection and deleted the hidden share specific .yml. 

Was having an issue getting comments to work testing the end to end using attack range.  The original search I was trying with comments was this and could use help fixing the syntax before the PR is merged:

'| `comment("GET WINDOWS EVENT LOGS THAT CAN CONTAIN NETWORK SHARE OPERATION INFO OVER DATE RANGE")`
      tstats `security_content_summariesonly` count values(Processes.user) as user values(Processes.parent_process)
        as parent_process min(_time) as firstTime max(_time) as lastTime from datamodel=Endpoint.Processes
        where (Processs.process_name=net.exe OR Processes.process_name=net1.exe) by
        Processes.process Processes.process_name Processes.dest | `comment("CLEAN UP OUTPUT, SEARCH ON KEYWORDS RELATED TO CREATE/DELETE EVENTS")`
        | `drop_dm_object_name(Processes)`
        | `security_content_ctime(firstTime)`| `security_content_ctime(lastTime)` | search (process=*share* OR process=*delete*)
        | `comment`("UNCOMMENT THIS REGEX TO SEARCH FOR ONLY EVENTS RELATED TO HIDDEN SHARES: regex process="\S+[$]")` | `create_or_delete_windows_shares_filter`'


macros/create_hidden_shares_filter.yml
```diff
@@ -0,0 +1,3 @@
+definition: search *
```
patel-bhavin: We dont neeed this macro file as its not being used by the detection search. Users should probably use the create_or_delete_windows_shares_filter macros to whitelist/blacklist

PR #324, opened 2020-01-27
update version info, add macros for first time seen cmd line detection

jzsplunk:
Detection worked out of the box in this case just updating the boiler plate.


patel-bhavin:
@jzsplunk : can you please update the modification_date of stories this detection belong to?

PR #327, opened 2020-01-27
updating boiler plate, detection not working out of box against T1193…

jzsplunk:
… but this may be due to error loading lookup table for supicious file extensions.


patel-bhavin:
update mod_dates for the stories this search belong to.

PR #383, opened 2020-03-02
CRL-1731: adding AND condition

patel-bhavin:


patel-bhavin:
Duplicate PR, Xilinx created a PR which was approved and merged!

PR #387, opened 2020-03-02
CRL-1735: adding escape chars

patel-bhavin:


rvaldez617:
Please check the backslashes. First line seems to have 2x as many as the others. That doesn't seem right. If you update, please also update version/modification_date and formatting as appropriate.

Issue #370, opened 2020-02-21
please add cloud as security_domain value


rosplk:
https://github.com/splunk/security-content/blob/c06702a7f4a9923e5105b1815e1cd6cc615cb857/spec/v2/detections.spec.json#L1464

patel-bhavin:
PR Created: https://github.com/splunk/security-content/pull/391

PR #391, opened 2020-03-02
Issue -370 spec update for security domain

patel-bhavin:
added cloud as a security_domain enum values


d1vious:
closes issue #370

PR #411, opened 2020-03-13
DONT MERGE! CRL-1747 updating all aws searches to use src_user

patel-bhavin:
Updated all detections and baseline searches to use a common field: src_user

Updated investigation search and deleted a duplicate


d1vious:
Hey we should close PR's that are not meant to be merged @patel-bhavin

PR #435, opened 2020-04-16
Updated ATT&CK Technique mapping to T1490

secsponge:


secsponge:
"Impact" wasn't in spec file so tests failed. Will fix and try again later

PR #439, opened 2020-04-17
points run story button to the ASX app 

d1vious:
points run story button to the ASX execute page with the story selected instead of github url looks sloppy.. 

When the app is not installed it displays the following error
image


d1vious:
tested this with the attack range, also updated the range with the latest build release of ASX

PR #456, opened 2020-05-18
Fixed typo in search, search optimization

zelienka:
Changes Processs to Processes.
Moved process search to tstats.


patel-bhavin:
Thanks for creating a PR with the fix! @zelienka

PR #457, opened 2020-05-18
Update attrib_to_hide_files.yml

jwindley-splunk:
Adding dest


patel-bhavin:
@jwindley-splunk : Thanks for creating a PR with the patch. Merging!

PR #459, opened 2020-05-19
Update create_or_delete_network_shares.yml

jwindley-splunk:
By having process=*share* AND process=*delete* you will miss any creation of shares. If you are just looking for creating OR deleting of shares you just need to look for the word *share*


patel-bhavin:
Nice catch, tested the updated search on attack_range using T1126 and it works fine

PR #462, opened 2020-05-19
Suspicious zoom process

trogdorsey:
Looking for first time seen child processes of zoom on both windows and macos.. Also added zoom.exe as a process that shouldn't spawn cmd.exe.

Update generate and spec to support collections as a type of lookup


patel-bhavin:
Closing this PR,  created a new PR with spec3 updates

PR #482, opened 2020-05-28
Zoom Story - converted to spec3 and other updates

patel-bhavin:
1. all manifests in spec3
2. updated the bin files to support collections.  Thanks to Dorsey
3. minor change in the Jinja files for baselines


patel-bhavin:
tested on a fresh attack_range splunk instance

PR #538, opened 2020-07-06
Testing detections

P4T12ICK:
Testing and Improvement of detections:
- Detect Path Interception By Creation Of program exe
- Windows Event Log Cleared
- Short Lived Windows Accounts
- Remote Desktop Network Traffic

You don't have to spend too much time for reviewing. The automated pipeline should catch the errors.


bin/ci-generate.py
```diff
@@ -12,31 +12,32 @@
...
-
-def load_file(file_path):
-    with open(file_path, 'r') as stream:
-        try:
-            file = list(yaml.safe_load_all(stream))[0]
-        except yaml.YAMLError as exc:
-            print(exc)
-            sys.exit("ERROR: reading {0}".format(file_path))
-    return file
+# def load_objects(file_path, VERBOSE):
```
jzsplunk, line 23: Are these load functions not needed any more?  Could we port them to a utility file for reuses in other parts of the app?

PR #542, opened 2020-07-08
detection testing: Detect New Local Admin account

P4T12ICK:
detection testing: Detect New Local Admin account


P4T12ICK:
merged into other PR

PR #557, opened 2020-07-12
Gcpk8plussearches

rosplk:
Added 6 GCP Searches for K8s sensitive role, object and kubectl supsicious calls


patel-bhavin:
@rosplk : there are a few mistakes in the PR which is why validate is failing:

1) Ensure that the manifest yml, search name and the output filter have the exact same name. the output filter is just all lower case of thee search name separated by _

2) search starting with sourcetype= <something> is not allowed, please replace 

sourcetype="google:gcp:pubsub:message" with `google_gcp_pubsub_message` ( a macro)

PR #575, opened 2020-07-22
[TR-037] test for process execution via WMI (

jzsplunk:
When running simulate I generated 2 rows of data but the test failed when I set count = 2 and passed when count = 1 minor detail that I was hoping you could comment on @P4T12ICK 

Screen Shot 2020-07-22 at 12 58 30 PM


P4T12ICK:
Hi Jo, this is one line of data, that's why count = 1 is correct. The header with the field names doesn't count. And it also doesn't matter if you already have a count > 1 before.

PR #576, opened 2020-07-23
Updated to use Type to define Baseline - Detection and Response Task Names

kfondren7:


d1vious:
@kfondren7 I reopened this as another PR since CI was failing PR #578

PR #578, opened 2020-07-23
Relative path and type consideration - kfondren7

d1vious:
Related to PR https://github.com/splunk/security-content/pull/576  from the kind work done by @kfondren7


patel-bhavin:
Thank you for the PR

PR #586, opened 2020-07-28
Windows DNS SIGRed CVE-2020-1350

davisshannon:
Properly validating submission now, sorry.

Includes detections for Zeek and Stream, macro yml files for stream_tcp and stream_dns, and the analytic story.


davisshannon:
Removing to add a new PR for another CVE

PR #605, opened 2020-08-02
Windows DNS SIGRed CVE-2020-1350 and F5 TMUI RCE CVE-2020-5902

davisshannon:
Detections, Stories, and Macros for CVE-2020-1350 and CVE-2020-5902

PR #608, opened 2020-08-04
F5 TMUI RCE CVE-2020-5902 - contribution from Shannon ymls and updates

patel-bhavin:


d1vious:
Originally from PR #605

PR #612, opened 2020-08-11
Layer 2 Cisco Network detections

inspired:
This detection works in conjunction with the Cisco Networks Add-on for Splunk and if used in conjunction with Cisco network devices configured with DHCP Snooping, will create a Notable Event when a Rogue DHCP server is handing out DHCP leases


inspired:
This detection works in conjunction with the Cisco Networks Add-on for Splunk and if used in conjunction with Cisco network devices configured with DHCP Snooping, will create a Notable Event when a Rogue DHCP server is handing out DHCP leases

d1vious:
@inspired this is wonderful thank you so much!

PR #613, opened 2020-08-12
converted lookup to collection

d1vious:
this was reported to have issues during bundle replication since the lookup gets really big.


patel-bhavin:
adding latestCount

PR #630, opened 2020-08-24
[TR-82] add in logic for risk tagging after synching with ESCU requirements

jzsplunk:
Updates to generate.py, validate.py and jinja files  for risk attribution
Sample yml update: detections/abnormally_high_aws_instances_launched_by_user.yml
detection spec update with examples


P4T12ICK:
I think there are a bunch of things missing:
- change of the spec file (not only the .md file)
- add some validation checks for this value if possible

Questions:
- are the risk values mandatory or optional? (I hope optional)
- what happens with existent content? Will they all get 3 fields more with default values or will dedicated RBA content developed which have these 3 fields?

d1vious:
@P4T12ICK some quick off the cuff answers for your questions:

 - are the risk values mandatory or optional? **optional**
 - what happens with existent content? **We will let the content author decide if he wants to use/include RBA fields when he writes the detection, not included by default. The goal is to enable anyone to tag the fields and modify the detection to be used by RBA, but not modify content to use it by default**

bin/validate.py
```diff
@@ -121,6 +121,28 @@ def validate_standard_fields(object, uuids):
...
+    # risk_object, risk_object_type and risk_score
+    # the first two fields risk_object, and risk_object_type are an enum of fixed values
+    # defined by ESCU risk scoring
+    if 'risk_object' in object:
+        try:
+            object['risk_object'].encode('ascii')
+        except UnicodeEncodeError:
+            errors.append("ERROR: description not ascii for object: %s" % object['name'])
+
+    if 'risk_object_type' in object:
```
patel-bhavin, line 134: we should enforce the risk_object_type to be limited to these three values: user, system, other

d1vious (dismissed):
Looks good to me!

d1vious (approved):

PR #693, opened 2020-09-30
fix workbench panel

P4T12ICK:
fixed workbench panel


patel-bhavin:
i removed an extra ] in the stanza name.

patel-bhavin (approved):

PR #715, opened 2020-10-08
Organize detections

d1vious:
places detections under folders for organization specifically:
* cloud
* network
* application
* endpoint
* web


miskoSplunk (approved):

ghoto (approved):
LGTM

bin/doc-gen.py
```diff
@@ -684,7 +684,13 @@ def parse_data_models_from_search(search):
     gmd = args.gen_markdown_docs
 
     stories = load_objects("stories/*.yml")
-    detections = load_objects("detections/*.yml")
+    detections = []
+    application = load_objects("detections/application/*.yml")
```
peter-cg, line 688: It may be easy to miss one of these files if we add a new category. Could also just `load_objects("detections/*/*.yml")` since we pass that to `glob.glob`.

bin/generate.py
```diff
@@ -569,22 +569,34 @@ def generate_mitre_lookup():
...
+    network = load_objects("detections/network/*.yml", VERBOSE)
+    endpoint = load_objects("detections/endpoint/*.yml", VERBOSE)
+    cloud = load_objects("detections/cloud/*.yml", VERBOSE)
+    detections = application + web + network + endpoint + cloud
+
     try:
         if VERBOSE:
             print("generating Mitre lookups")
-        generate_mitre_lookup()
+        #generate_mitre_lookup()
```
peter-cg, line 588: Did you mean to disable this?

d1vious: no good catch fixed!

bin/doc-gen.py
d1vious: agree, much cleaner way to do this, fixed in all bin scripts I touched. Thank you!

d1vious:
@peter-cg just implemented the fixes you recommend, let me know if this is ok to merge.

P4T12ICK (approved):
looks good to me

PR #804, opened 2020-10-19
Updating credential dumping notebooks

peter-cg:
I verified that both notebooks still fire on the test data.


miskoSplunk:
Seems good to me

miskoSplunk:
Seems good to me!

miskoSplunk (dismissed):
Finally found a place where it's not only a comment, but has Approve checkbox ...
Approving!

miskoSplunk (approved):

PR #810, opened 2020-10-21
Data Exfil PR

davisshannon:
Generic Data Exfiltration Analytics Story, along with a detection for the recently released POC code for SNICat (exfil over TLS SNI).  Macros included.


d1vious (changes requested):
@davisshannon hey this is awesome, we need a few changes though before we can merge in! I would rename the story title to use camel casing eg: `Data Exfiltration`. Also on the detection I would include a tailing `| table` command that displays to the user the data he would care about for the results, think of the fields a user would pass to a response playbook. Finally we need (starting to require) a data set and a test file for this detection. Check out this one for example: https://github.com/splunk/security-content/blob/new_test_file_format/tests/endpoint/creation_of_shadow_copy.test.yml

josehelps:
Also looks like CI failed with the following error: 
```
processing manifest ./stories/data_exfiltration.yml
while scanning a simple key
  in "./stories/data_exfiltration.yml", line 7, column 1
could not find expected ':'
  in "./stories/data_exfiltration.yml", line 8, column 1
Error reading ./stories/data_exfiltration.yml
```

I do not think the story is valid yml. https://app.circleci.com/pipelines/github/splunk/security-content/3515/workflows/04077230-c17c-4d1b-92d4-ab9129a7536d/jobs/10638

PR #813, opened 2020-10-21
First test end to end

ghoto:
This is WIP, don't merge yet.

I've added a script that does end to end testing of a SSA detection. In order:
- read testing file
- download test data locally
- extract SPL2
- replaces SSA macros
- downloads latest splunk-streaming-ml with humvee
- executes the SPL2 query locally and outputs the results

The results can then be used to see whether the detection runs correctly.

Example:
```
python bin/validate_ssa.py tests/endpoint/*
```


d1vious (changes requested):
I would remove the validate.py execution since this is a duplicate from circleci, I just added it for testing https://github.com/splunk/security-content/pull/813/files#diff-037ea159eb0a7cb0ac23b851e66bee30fb838ee8d0d99fa331a1ba65283d37f7R13

bin/validate_ssa.py
```diff
@@ -0,0 +1,103 @@
...
+        subprocess.run(["java",
+                        "-jar", get_path("%s/humvee/build/libs/humvee-1.2.1-SNAPSHOT-all.jar" % SSML_CWD),
+                        'cli',
+                        '-i', spl2_file,
+                        '-o', test_out],
+                       stderr=subprocess.DEVNULL)
+        with open(test_status, "r") as test_status_fh:
+            res = '\n'.join(test_status_fh.readlines())
+            if res == "OK\n":
+                print("SPL2 executed correctly")
```
josehelps: have a explict exit(0) if successfuly

ghoto:
This is the current state of the branch. I would say we could perhaps merge despite there are 2 tests failing. 
system_process_running_unexpected_location___ssa.test.yml fails because it doesn't have any data to test.
tests/application/detect_phishing_content___ssa.test.yml supposedly TR-162 will fix this, as there are fields that contains ":" and "`" to be removed.

In this commit I had to fix tests pass conditions, and fix also the datasets in the s3 bucket that didn't have the right format (missing fields used by the detection, timestamp will come as string from the ssa-parser and not a number)

```
Passed tests
=================
tests/endpoint/attempted_credential_dump_from_registry_via_reg_exe___ssa.test.yml
tests/endpoint/detect_dump_lsass_memory_using_comsvcs___ssa.test.yml
tests/endpoint/detect_prohibited_applications_spawning_cmd_exe.test.yml
tests/endpoint/first_time_seen_cmd_line___ssa.test.yml
tests/endpoint/prohibited_apps_spawning_cmdprompt___ssa.test.yml
tests/endpoint/rare_parent_process_relationship_lolbas___ssa.test.yaml
tests/endpoint/unusual_lolbas_in_short_period_of_time___ssa.test.yml
tests/endpoint/unusually_long_command_line___ssa.test.yml

Failed tests
=================
tests/application/detect_phishing_content___ssa.test.yml
tests/endpoint/system_process_running_unexpected_location___ssa.test.yml
```

peter-cg:
This is really cool. Looks like the gitlab CI job fails because it can't find java -- might work better to run it via gradlew? The docs mention you can do `./gradlew humvee:run --args="cli -i input.spl2 -o results.out"`

Edit: I see you just removed the repo, so maybe just try `$JAVA_HOME/bin/java`

ghoto:
> This is really cool. Looks like the gitlab CI job fails because it can't find java -- might work better to run it via gradlew? The docs mention you can do `./gradlew humvee:run --args="cli -i input.spl2 -o results.out"`
> 
> Edit: I see you just removed the repo, so maybe just try `$JAVA_HOME/bin/java`

yes, I'm aware of this. So I don't understand 100% how the ci thing works in gitlab. Jose told me that the image might not have java, so I'm changing the image to something that has java-8 and python3

I moved away from the gradlew option because I don't want to clone and build anymore. We need credentials to pull the code from streaming-ml. It works locally though.. Right now the best option is to just download the jar directly from the repo (works locally, and on gitlab).

Another option could be to use their docker image to run humvee, but it seems it doesn't contain the jar, neither python3

.gitlab-ci.yml
```diff
@@ -6,10 +6,9 @@ stages:
 
 validate:
   stage: ssa-validate
-  image: python:3.7
+  image: chrsep/alpine-java-python
   before_script:
     - pip3 install -r requirements.txt
   script:
     - python3 bin/validate.py -p . -v
```
peter-cg: I think Jose wanted you to remove the non-SSA validate since that's done elsewhere.

peter-cg (approved):
Looks like the image change did the trick -- nice!

d1vious (approved):
looks good

PR #816, opened 2020-10-21
Credential Extraction detected via common exploit modules ...

miskoSplunk:
… embedded in Mimikatz, PowerSploit and DSInternals


josehelps:
@miskoSplunk can we include the test files and data sets for each of these? For example: https://github.com/splunk/security-content/blob/develop/tests/endpoint/attempted_credential_dump_from_registry_via_reg_exe___ssa.test.yml. Also please rename all files to use ssa___* as a prefix and not a suffix (I would like to start changing this across our current content). Also rename the notebook to be all under score and have a prefix of ssa___. Thank you!

d1vious (changes requested):
see https://github.com/splunk/security-content/pull/816#issuecomment-717346878

PR #823, opened 2020-10-23
adding smoke test for master api

josehelps:
This avoids us pushing changes without making sure that the API is responding with 200 OK after a push. The CI job should catch this if it does happen.


.circleci/config.yml
```diff
@@ -384,6 +384,17 @@ jobs:
             aws s3 cp lookups s3://security-content/lookups --recursive --exclude "*" --include "*.csv"
             aws s3 cp macros s3://security-content/macros --recursive --exclude "*" --include "*.yml"
             aws s3 cp deployments s3://security-content/deployments --recursive --exclude "*" --include "*.yml"
+      - run:
+          name: security content api smoke test
```
peter-cg, line 354: How quickly does the endpoint pull the updated content? (Should we add a delay here between updating and checking?)

josehelps: its instance, the API reads from that bucket every time a request is made. This is not a polling/pulling operation, it is the backend DB for it basically 😄

PR #838, opened 2020-10-26
fixed test and detection - unusual location

ghoto:
Current detection had issues with very long WHERE clause.

This PR solves this issue by breaking this clause into multiple ones, and performing a union at the end. Tests were also failing because of this very long where conditions, taking too long to finish and exhausting the resources.

+ Added dataset for testing
+ Modified detection by splitting and union. Performance improved in terms of execution time.


josehelps (approved):
this looks good, although breaking things into conditions looks super messy, do we know exactly when this will be fixed?

PR #843, opened 2020-10-27
Ransomwares3

rosplk:
Searches based on KMS Encrypt policy and S3 bucket object encryption


miskoSplunk:
Seems that detections are reasonable, if a customer does follow use of Multi Factor Authentications (MFA) on all accounts as suggested in https://docs.aws.amazon.com/IAM/latest/UserGuide/data-protection.html.

However, could we extend this with a "lookup"  to check for cases where customer has a lot of non-MFA accounts - and then not run this, but instead warn that MFA should be used on all accounts?

detections/cloud/aws_detect_users_creating_keys_with_encrypt_policy_without_mfa.yml
```diff
@@ -0,0 +1,22 @@
+author: "Rod Soto, Splunk"
+date: "2020-10-27"
+description: "This search provides detection of accounts creating KMS with encrypt policy and without using MFA."
+how_to_implement: "You must install splunk AWS add on and Splunk App for AWS. This search works with clodtrail logs"
+id: c79c164f-4b21-4847-98f9-cf6a9f49179e
+known_false_positives: "Not all KMS key creations are malicious. Particulary look for kms:Encrypt only policy. It is very unusual to create a key to encrypt only and not to decrypt."
+name: "aws detect users creating keys with encrypt policy without mfa"
```
patel-bhavin: @rosplk : Can you update all the search names to use upper case for the first letter of every word?

detections/cloud/aws_detect_users_creating_keys_with_encrypt_policy_without_mfa.yml
```diff
@@ -0,0 +1,22 @@
...
+references:
+   - https://rhinosecuritylabs.com/aws/s3-ransomware-part-1-attack-vector/
+   - https://github.com/d1vious/git-wild-hunt
+   - https://www.youtube.com/watch?v=PgzNib37g0M
+search: 'sourcetype:aws:cloudtrail eventName=CreateKey userIdentity.sessionContext.attributes.mfaAuthenticated=false | spath Parameters.policy=*kms.Encrypt*  |table userIdentity.sessionContext.attributes.creationDate userIdentity.sessionContext.attributes.mfaAuthenticated user user_access_key requestParameters.keyUsage responseElements.keyMetadata.arn
```
patel-bhavin: searches shouldn't start with "sourcetype:aws:cloudtrail ".  Use a macro instead"`sourcetype:aws:cloudtrail`"

patel-bhavin (changes requested):

detections/cloud/aws_detect_users_creating_keys_with_encrypt_policy_without_mfa.yml
jzsplunk: @patel-bhavin is there a SPL style guide (like PEP 8 for python) or reference we can use when checking best practices for a particular searches readability?

patel-bhavin: yeah, there is an old one which I compiled but it needs a refresher:  https://confluence.splunk.com/display/MG/ESCU+Best+Practices

patel-bhavin: also, after doing some testing: looks like the search syntax and logic is in correct.

patel-bhavin:
@rosplk : Finally, for the second search
image

PR #877, opened 2020-11-04
Applying and setting stolen credentials - detections via …

miskoSplunk:
…  PowerSploit, Mimikatz and DSInternals APIs


detections/endpoint/ssa___setting_credentials_via_powersploit_modules.yml
```diff
@@ -0,0 +1,45 @@
...
+version: 1
+date: '2020-11-03'
+description: This detection identifies illegal setting of credentials via PowerSploit modules. 
+how_to_implement: You must be ingesting Windows Security logs from devices of interest, including the event ID 4688 with enabled command line logging.
+references: 
+- https://github.com/PowerShellMafia/PowerSploit
+type: SSA
+author: Stanislav Miskovic, Splunk
+search: '| from read_ssa_enriched_events()
+
```
jzsplunk, line 12: minor detail @miskoSplunk  whitespace on lines is for readability in this case for the SPLv2?

detections/endpoint/ssa___applying_stolen_credentials_via_powersploit_modules.yml
```diff
@@ -0,0 +1,59 @@
...
+type: SSA
+author: Stanislav Miskovic, Splunk
+search: '| from read_ssa_enriched_events()
+
+| eval timestamp=parse_long(ucast(map_get(input_event, "_time"), "string", null)),
+       cmd_line=ucast(map_get(input_event, "process"), "string", null)
+| where cmd_line != null AND
+        ( 
+           match_regex(cmd_line, /(?i)Invoke-CredentialInjection/)=true OR
+           match_regex(cmd_line, /(?i)Invoke-TokenManipulation/)=true OR
```
jzsplunk, line 19: Maybe an improvement for the case of (?i) using the regex operator is to keep a list of tokens per string you want to match on (for instance (invoke,credential,injection) and use longest common subsequence for more fuzzy matching or partial matching...

jzsplunk (dismissed):

detections/endpoint/ssa___setting_credentials_via_powersploit_modules.yml
miskoSplunk: Yes, not necessary, but I left blank lines so a reviewer can easily see the detection - which is extraction of fields that will be used and the where condition ... the rest is just a loading/reporting overhead of SSA.

detections/endpoint/ssa___applying_stolen_credentials_via_powersploit_modules.yml
miskoSplunk: For processing efficiency, I agree that specifying just few key words would speed up regex matching.

However, here I must look for the exact names of exploit kit modules (such as Invoke-CredentialInjection) vs. just looking for "credential" and "injection" terms. Reason being "false positives", i.e., otherwise I'd increase risk of alerting on legitimate powershell modules used by admins.

jzsplunk (approved):

PR #883, opened 2020-11-05
WIP: Downloads latest Humvee with DSP SCS and streaming-ml

ghoto:
Fetches gradle.properties from streaming-ml repository to get all the latest versions.
Fetches the latest humvee.jar from artifactory that matches this version.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,7 +8,7 @@ validate:
   stage: ssa-validate
   image: docker-hub.repo.splunkdev.net/alpine:3.7
   before_script:
-    - apk add --update python3 python3-dev py-pip build-base openjdk8-jre && pip install virtualenv && rm -rf /var/cache/apk/*
+    - apk add --update python3 python3-dev py-pip build-base openjdk8-jre && pip install virtualenv git && rm -rf /var/cache/apk/*
     - pip3 install -r requirements.txt
   script:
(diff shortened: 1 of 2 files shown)


jzsplunk (approved):
This looks great @ghoto. 

 Random question we can take offline:  I am a bit confused at internal use of scala dependencies and the structure of Humvee if I look at this from outside of Splunk org. Does humvee have an equivalent to something that is not 'paywalled' that a customer could use if they don't have access to ssa in terms of testing/modifying SSA specific content down the road?

jzsplunk (dismissed):

ghoto:
Please don't merge this. I can't pull the gradle properties file from their repo inside docker since it lacks credentials. I still need to think what would be the best way to get the latest version.

peter-cg:
Instead of grabbing the repo config it's probably better to get the version info directly from artifactory (since that's what is actually published). There's a `latestVersion` REST endpoint but that doesn't seem to work for humvee (probably some artifactory config issue).

But you can just
`GET https://repo.splunk.com/artifactory/api/search/artifact?name=humvee&repos=maven-splunk-local`
and filter for the uri containing `com/splunk/humvee-scala_2.11/latest`.

ghoto:
This PR is obsolete. There is a new one, and will delete the branch TR-254

PR #894, opened 2020-11-06
specific token that requires github repo access otherwise auto merges…

josehelps:
… fail


josehelps:
Noticed that github-action bot is not allowed to merge by dependabot see: https://github.com/splunk/security-content/pull/887/#issuecomment-722398403 this should resolve this issue.

ghoto (approved):
LGTM

PR #926, opened 2020-11-13
Find problems with declared SSA fields in detection file

ghoto:
What's in this PR
============

Refactoring of validating/testing (both are separate scripts)
- Validation inspects the detection description tag. Syntax checks, missing tags, etc.
- Testing executes unit tests and validates output

What's being validated for SSA detections:
- risk_severity is populated
- input and output fields required for SSA events
- required fields tag matches search


josehelps (changes requested):
I would rename bin_tools, to modules or lib, also rename scripts to start with ssa_validate.py and ssa_test.py. We should also consider moving the testing service as a single command you can kick off here similar to escu_validate.py and escu_test.py what do you think @P4T12ICK?

josehelps (approved):

PR #936, opened 2020-11-17
removing a broken detection

josehelps:
tested it in attack_range_local with windows 10. It does should up due to ansible:

image


P4T12ICK (approved):
I think that is the right decision

PR #945, opened 2020-11-20
Refactored and pulls data only on ssa detections

ghoto:
This refactoring is to solve the problem that ssa_test.py downloads test dataset on tests that are not SSA. It download data once detects that the search is SSA.


ghoto:
@P4T12ICK let me know if this patch solves your problem.

This is a patch. We still need to figure out what will be the best way to proceed when we declare tests and their datasets.

P4T12ICK (approved):
Looks good to me

PR #959, opened 2020-11-24
Basic assertion language

ghoto:
- More flexible tests for SSA detections
- Make tests easy to write
- Keeping tests easy to be understood by a reviewer

I have defined a simple language where you can use logical operators (&,^,|,!) and functions "@function_name". To each function, we internally pass the output of the pipeline. For instance the toy test below tests that the number of events is between 0 and 10.

```
pass_condition: "@count_gt(0) & @count_lt(10)"
```

Functions can be added to global_assertions file, and for specific custom assertion functions, those can be added to it's detection category. For instance you could add a function to modules.assertions.network.assertions that checks that events belong to HTTP flows, if it's that what you want to test.


P4T12ICK (approved):
Looks good to me. I would wait for the CI fix before merging it. Great work.

ghoto:
I'm updating branch. When it's done I'm gonna merge.

PR #962, opened 2020-11-25
Fixed initialization logic

dp-42:
Hello,

my colleague and I have noticed that there is an issue with the initialization logic when converting the detections to a story.

Lets assume you have a detection with 3 CIS tags.
**On initialization**, the current logic loops through all 3 tags and always sets the set to the current tag.
Therefore, when processing 3 CIS tags, it will only set the last tag and not all 3 tags.

Hope that helps!


P4T12ICK:
Thank you very much. We will test it.

patel-bhavin:
@dp-42 : We have tested and the PR looks good. Thank you so much for this contribution and sorry for the delay in testing this.

patel-bhavin:
Created a duplicate PR for testing.

PR #986, opened 2020-12-02
Changed test script and detections to comply with new format

ghoto:
Attack data is now part of the test. Multiple test eventually can be added to one single testing file.


P4T12ICK (approved):
Looks good to me

PR #987, opened 2020-12-03
Remove wrong workbench panels

patel-bhavin:
These two changes are merged in develop already 

1) Removed workbench investigation panels that are already shipped by ES
2) Updating jinja/response tasks files to test workbench with multiple tokens


patel-bhavin:
These two changes are merged in develop already

Removed workbench investigation panels that are already shipped by ES
Updating jinja/response tasks files to test workbench with multiple tokens

PR #1000, opened 2020-12-08
Update appinspect to use API

josehelps:
Today appinspect checks use the prebuilt CLI this is not update with the latest checks also it does not return a failure of the CI job if it fails. This fixes the failure status issue as well as leverages the API. 

Making a bunch of fixes for the app-inpect failures


P4T12ICK:
Hey guys, curious why in this branch are so many changes to response_tasks and macros? Ans also some work on workbench panels. I just want to be sure that we not delete something important.

patel-bhavin (dismissed):

patel-bhavin (approved):

PR #1034, opened 2020-12-16
o365rodetections

rosplk:
first detection


miskoSplunk:
Let's add test datasets too.

PR #1042, opened 2020-12-20
First push of new content for Phantom

socologize:
This first pull of content has the Indicator Enrichment objects and the new response plans being developed for my principal project. - socologize (Rob Gresham)


socologize:
closed the pull request to due to two issues with playbook folders and spec schema for the response_tasks.

PR #1098, opened 2021-01-14
Ransomware fixups

splunkins:
Tidy-up of the ransomware extensions csv (removing entries with no file extensions), and a reconfiguration of the lookup to be case-insensitive (for better matching of mixed-case file extensions).


patel-bhavin:
thank you for your contribution and  fixing the empty lookups, since we had some CI changes i created a new PR with your changes: https://github.com/splunk/security-content/pull/1120

PR #1125, opened 2021-01-22
Name changes

trogdorsey:
Changed some language in a bunch of files


miskoSplunk:
I'm ok with these changes.

Good refresher on Splunk-preferred ways of referring to filtering: block lists and add lists.

patel-bhavin (approved):

PR #1137, opened 2021-01-27
Mustang investigations

patel-bhavin:
For bug reported : https://jira.splunk.com/browse/TR-395 and https://jira.splunk.com/browse/TR-396

 updating all necessary `jinja` and and `generate.py` for ES use-case investigations for a story
 Workbench panel updates
 Tested on Attack_range


P4T12ICK (approved):
Looks good to me

PR #1141, opened 2021-01-27
Baron Samedit Detection, Story, Macros

davisshannon:
Detections for https://blog.qualys.com/vulnerabilities-research/2021/01/26/cve-2021-3156-heap-based-buffer-overflow-in-sudo-baron-samedit

PR #1153, opened 2021-01-29
Baron Samedit Detection, Story, Macros - duplicating files and experimental

patel-bhavin:
Duplicating Files from shannons PR https://github.com/splunk/security_content/pull/1141/files since app-inspect was failing
- moving searches to experimental until we figure out a better way to test syslog and TA-Osquery datasets


josehelps:
@patel-bhavin do we need a linux endpoint on the attack_range with osquery support for this to be tested?

MHaggis (approved):

patel-bhavin:
@d1vious yes, an endpoint would be great but as long as we have the TA and the dataset, we can test these searches

PR #1164, opened 2021-02-01
Fixed SSA with cmd line norm

ghoto:
Normalization of command lines for the sake of finding first times, while preserving original cmd line.
- Paths in CMD line are converted to PATH
- Numbers in CMD line are converted to N

E.g. cmd.exe /c command.exe ./temp/1234 => cmd.exe /c command.exe PATH
cmd.exe /c command.exe -ts 16002991002 => cmd.exe /c command.exe -ts N

first_time_seen_command_line_argument.yml perhaps need to change too.


miskoSplunk:
@ghoto could you please explain why would we want/need to "featurize"/generalize command lines?

If this is to suppress a long tail resulting from command-line arguments, 
then it's likely we'd need a more comprehensive model instead of trying to tweak this simple one. 

Here is a thought, if we just turn things to say a "PATH" to track only generic forms of calling cmd.exe; then, we may loose a lot of useful detections - e.g., if you do cmd.exe $SPLUNK_PATH/splunk start - that's nothing suspicious. However, if you run cmd.exe ./regedit, it's much more suspicious.

Also, how would we explain this "featurized" process call detection: "first time a process was called with a 'strange' syntax"?

ghoto:
@miskoSplunk These are good points. Perhaps we can add these corner cases to the unit test. We should perhaps to improve the regex to handle better the case you have [./executable.exe => PATH/executable.exe] instead of [./executable.exe => PATH].

The original assumption of the detection (ESCU) was that the number of commands written directly from "cmd.exe /c" were kind of limited, so we didn't care about things could change in the command line.

Now the problem is that in reality a command can be pointing to some temporal folder, or it could contain other artifacts that wouldn't change the meaning of the execution.

harsha2010:
@ghoto where are the FPs coming from? do you have a dataset with an analysis i can look at?

patel-bhavin (approved):

PR #1168, opened 2021-02-02
TR-380 change type field 

josehelps:
This PR adjusts all the scripts, spec, documentation and detections to use instead of use the `type` field to describe the product eg (ESCU, SSA, etc.) instead changes it for the type of search, eg `batch` or `streaming`. Products will be named under a new tag field called `products`. Also introduced a new type called `response` under anything related to responses.


P4T12ICK:
@d1vious you probably used a script to convert the existing detections. I think you forgot to give an order into the yml dump. Therefore all keys are sorted alphabetical, which is somehow weird when the name of the detection is somehow in the middle. I think we should keep our order as we have it today. 

I had a look into my old scrripts:
you first assign the values into a dict in that order which you want to have it in the file and then:
`            with open('macros/' + new_file_name + '.yml', 'w+' ) as outfile:
    	           yaml.dump( new_dict , outfile , default_flow_style=False, sort_keys=False)
`

P4T12ICK:
We also need to update the GitLab CI jobs like ssa_test and ssa_validate.

P4T12ICK:
Let me know when you run your script then I will do some testing by deploying it to a Splunk instance.

d1vious:
@P4T12ICK Fix in https://github.com/splunk/security_content/pull/1185

PR #1174, opened 2021-02-04
Create msbuild_spawn_via_office_product.yml

MHaggis:
Analytic to capture an office product spawning MSBuild


MHaggis:
merging with other release

PR #1178, opened 2021-02-08
TR-410 rename analytics_story to analytic_story

josehelps:
renames  `analytics_story` to `analytic_story` everywhere.


MHaggis (dismissed):

P4T12ICK:
Here we have the same problem of ordering of the detections. Maybe that was already introduced through the type change not sure. I suggest to use the old ordering of the keys. See my comment in the other PR for the code snippet.

josehelps:
duplicate since it was merged in another PR.

PR #1185, opened 2021-02-11
TR-377 add datamodel field

josehelps:
- [x] adds a datamodels as an array field across all detections
- [x] updates detection spec to include datamodel field
- [x] add new tool `pretty_yaml.py` that organizes all detection keys in a specific order
- [x] add CI step in the update_github_sources job to run pretty_yaml.py.

Already includes PR #1168 PR #1178


P4T12ICK (dismissed):
Looks good to me. I also tested the ESCU build. One minor thing what could be done is also applying an order insides tags.

MHaggis (dismissed):

MHaggis (approved):

PR #1209, opened 2021-02-24
updating path for detection service

patel-bhavin:
This PR updates the `automated_detection_testing/detection_service.py` to use contentctl


P4T12ICK:
@patel-bhavin why is that needed? I don't see the need to change it.

P4T12ICK (approved):
Looks good to me

PR #1215, opened 2021-02-25
Awsiamsearches

rosplk:
AWS IAM discovery searches


patel-bhavin:
@rosplk : the response tasks yml have a different spec: for example https://github.com/splunk/security_content/blob/develop/response_tasks/get_all_aws_activity_from_country.yml

in the spl itself: add a token like `user =$user$`, so we have use these searchesfor responses

josehelps:
depends on an update to the response spec. Let me focus on this before we merge this in.

patel-bhavin:
Closing this PR for now, we need to move these into response_tasks when we get the playbooks working

PR #1216, opened 2021-02-26
first workbook example

philroyer-phantom:
Add a single workbook and all of its phases and tasks to demonstrate
our new format for workbooks in security content.

Good examples with all the fields are the workbook "Endpoint Ransomware", the phase
"Contain", and the task "Neutralize Propagation Vectors".


josehelps:
Shelving this PR for now, since we will come back to it after we have migrated all playbooks.

PR #1217, opened 2021-03-01
update generator automated_detection_testing

patel-bhavin:
Tested and update docker file on AWS.

### DIFF
--- a/automated_detection_testing/detection_service.py
+++ b/automated_detection_testing/detection_service.py
@@ -81,7 +81,6 @@ def main(args):
         ssh_key.write(response['KeyMaterial'])
     os.chmod(ssh_key_name, 0o600)
     private_key_path = str(os.getcwd() + "/" + ssh_key_name)
-    print (private_key_path)
 
     # build new version of ESCU
     sys.path.append(os.path.join(os.getcwd(),'security_content/bin'))
@@ -241,4 +240,5 @@ def get_secret(secret_name):
 
 
 if __name__ == "__main__":
+
     main(sys.argv[1:])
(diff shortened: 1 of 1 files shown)


automated_detection_testing/detection_service.py
```diff
@@ -79,13 +80,14 @@ def main(args):
...
+    private_key_path = str(os.getcwd() + "/" + ssh_key_name)
 
     # build new version of ESCU
     sys.path.append(os.path.join(os.getcwd(),'security_content/bin'))
+
     try:
         module = __import__('generate')
-        module.sys.argv = ['generate', '-p', 'security_content', '-o' 'security_content/package']
-        results = module.main(module.sys.argv)
+        results = module.main(REPO_PATH = 'security_content' , OUTPUT_PATH = 'security_content/package', VERBOSE = '--verbose' )
```
P4T12ICK: @patel-bhavin, I think the correct value for VERBOSe is either True or False. Maybe we should go for TRUE.

P4T12ICK (changes requested):

automated_detection_testing/detection_service.py
patel-bhavin: Updated! Thanks for reviewing this

P4T12ICK: Another small thing: The True needs to be without ' otherwise it's a string and not a boolean.

patel-bhavin: hmmn, weird. it still worked for me and gives a verbose output. anyway, i updated it to boolen

P4T12ICK (approved):

PR #1221, opened 2021-03-03
line 1048

l-kutch:
At line 1048, there is a closing tag of </ol> that possibly doesn't have a corresponding opening tag. It's causing some syntax problems in https://docs.splunk.com/Documentation/ESSOC/3.15.0/stories/UseCase#Phishing_Payloads

"Powershell executes a reverse shell, rendering the exploit successful </ol>" 

I've proposed adding an opening tag, so that I can save in ponydocs but that is not quite correct syntax. Thanks.


patel-bhavin:
Hello @l-kutch : These docs are automatically created by our back end scripts `doc-gen.py`,  so we'd need to fix this in our script and not in this file `docs/splunk_docs_categories.wiki`. 

@d1vious : is currently working on re vamping the doc-gen script and some of these formatting errors will/should be taken care off in there. 

Closing this PR for now. Thank you for taking the effort to fork and create this PR for now.

PR #1242, opened 2021-03-11
Tr 381 add required fields

P4T12ICK:
Add required fields to detections


patel-bhavin:
added security_domain field to a few files, so that validate passes

patel-bhavin:
The PR looks good to me otherwise! awesome work homie

josehelps (dismissed):

PR #1283, opened 2021-03-19
Deprecate rare parent process ssa

josehelps:
this is something that is generating FPs we should revisit its design


ghoto:
I wonder why this hasn't trigger the CI pipeline in gitlab.

P4T12ICK:
duplicate of already merged PR

PR #1285, opened 2021-03-20
Improve validation script

Skypex:
Hello,

this PR implements the following changes:

- Fix a bug in the error aggregation logic
- Introduce improved `jsonschema` validation
- Update `.gitignore` to exclude VSCode configuration

The bug was caused by saving errors returned by the `validate_lookups_content(...)` function without aggregating them to the `errors` array. That means that errors have been overwritten while iterating through the lookups causing potential lost of errors.

The `jsonschema` part does essentially the same thing as before, but if an objects now contains multiple schema violations all errors will be returned. The original implementation would only return the first schema error of the tested object. This should improve usability.

Please have a look if the introduced changes make sense to you! 😄 

Kind regards,
Daniel


josehelps:
@Skypex this looks awesome thank you!

josehelps (approved):

PR #1287, opened 2021-03-22
TR-489 Update docgen

josehelps:
- [x] refactors doc_gen code to print new format used by docs team
- [x] adds the ability to generate docs for detections in markdown as well as wiki markup format
- [x] updates CI job to run doc gen on every PR
- [x] includes the ability to generate docs on every release for spec files.


P4T12ICK (approved):
looks good to me

Issue #1297, opened 2021-03-26
bug in how split mitre_ids they should be an array


josehelps:
https://github.com/splunk/security_content/blob/6212932c3573345c3e7714ae07cb35693d01ae76/bin/newcontent.py#L198

PR #1298, opened 2021-03-26
bug in mitre_id that caused it to not be stored as a yaml array 

josehelps:
This was the core bug that caused issue #1297


P4T12ICK (approved):
Looks good to me

PR #1299, opened 2021-03-26
Ssa smoketest nightly

P4T12ICK:
most of the ssa detections test moved to raw data


tccontre (approved):
as per request.looks good!

PR #1300, opened 2021-03-26
Fix finding testing file search in the right path

ghoto:
The current github helper doesn't check if .yml test file exists in the right path. 
git diff returns paths relative to the cloned directory (security_content), but the script's working directory is in ../security_content


P4T12ICK:
I'm not sure if we should do it. When we allow that people use *.yaml file then our content would break also at a lot of other places.

PR #1317, opened 2021-04-07
Tr 365 loblas first time seen dll logic 

jzsplunk:
I opened a seperate ticket for a small bug in ssa_validate.py that https://github.com/splunk/security_content/pull/1316 that was related to the SPLV2 logic in this detection not passing the validate logic.  This seems to be an issue with the validate script itself and not any of the SSA detections including the first time dll use case.  I am opening a PR despite the validate logic locally failing on all detections when we fix the validate syntax bug so for now keeping the bug fix in the separate branch TR-563-ssa-validate-fix.


detections/endpoint/ssa___first_time_seen_dlls_lolbas.yml
```diff
@@ -0,0 +1,37 @@
...
+date: '2021-3-1'
+author: 'Joseph Zadeh, Splunk'
+type: streaming
+datamodel: []
+description: Identify dlls that have been observed for the first time
+  in windows active directory logs.
+search: '| from read_ssa_enriched_events()
+         | eval timestamp=parse_long(ucast(map_get(input_event,"_time"), "string", null))
+         | eval process_name=lower(ucast(map_get(input_event, "process_name"), "string", null)), dest_user_id=ucast(map_get(input_event, "dest_user_id"), "string", null), dest_device_id=ucast(map_get(input_event, "dest_device_id"), "string", null)
+         | select process_name, dest_user_id, dest_device_id, timestamp
```
miskoSplunk: @jzsplunk please make sure about a couple of things:
1. we'd want to look only for a subset of process events that are about DLLs  (don't see any DLL spotting filters in the detection)
2. More likely than not, we'd be looking for DLLs in the "process" field, instead of looking for them in the "process name" field. For example, in example events in Jira (https://jira.splunk.com/browse/TR-365), all DLLs are in the "Process Command Line" field of raw events, which maps to the "process" field in SSA. To that end, we may need to do some additional parsing/extraction in the "process" field, to get the DLL ... see if we see it for the firs time in general, for a give user, or for the process that is trying to run it (we also need to get that process either from the "process name" field, or try to parse it from the "process" field.

jzsplunk:
@miskoSplunk I added the check for dll's and changed the field to process instead of process name that we evaluate?  What do you think about merging

detections/endpoint/ssa___first_time_seen_dlls_lolbas.yml
```diff
@@ -0,0 +1,39 @@
...
+datamodel: []
+description: Identify dlls that have been observed for the first time
+  in windows active directory logs.
+search: '| from read_ssa_enriched_events()
+         | eval timestamp=ucast(map_get(input_event, "_time"), "long", null),
+           process=lower(ucast(map_get(input_event, "process"), "string", null)),
+            --for now ignore the "process" field since this contains both process_name plus command arguments
+           dest_user_id=ucast(map_get(input_event, "dest_user_id"), "string", null),
+           dest_device_id=ucast(map_get(input_event, "dest_device_id"), "string", null)
+         | where like(process, "%.dll")
```
miskoSplunk, line 16: This is really not what we were looking for - it would match any command lines that have ".dll" in them.

Instead, this was the plan as per Jira TR-365 (https://jira.splunk.com/browse/TR-365) :
* Build a detection that identifies first times when a specific DLL is run.
* Consider looking for global/per-entire-customer-org first times vs. doing first times for a user or a device.
* Consider doing (another) simpler rule-based detection for DLLs that are not run from system folders.

I'll post some detection code in Jira, let's make datasets for it, and start from there.

miskoSplunk (changes requested):

jzsplunk:
This will be replaced by TR-605 branch logic and will not be merged closing the PR.

PR #1321, opened 2021-04-08
added new detection

rbecwar:
this is base on previous work


detections/endpoint/malicious_powershell_executed_as_a_service.yml
```diff
@@ -0,0 +1,47 @@
...
+id: 8e204dfd-cae0-4ea8-a61d-e972a1ff2ff8
+version: 1
+date: '2021-04-07'
+author: Ryan Becwar
+type: batch
+datamodel:
+- Endpoint
+description: This detection is to identify the abuse the Windows SC.exe to
+ execute malicious commands or payloads via PowerShell. 
+search: ' `wineventlog_system` EventCode=7045 Service_File_Name IN ("*powershell.exe*", "*-nop*", "*hid*")
```
miskoSplunk, line 11: @P4T12ICK could you please modify this a bit ... as it is, it would trigger even if this is not about powershell - for example, if we just find "hid" in the command line.
 
These are some suggestions:

* I would add some other names of the powershell executable: psexec, powershell_ise, and pwsh.exe (https://www.powershelladmin.com/wiki/PowerShell_Executables_File_System_Locations)

* Let's strengthen the "hide command" option, such that we add spaces and do some ORs: "* -nop *" OR "* -nopr *" OR ..... OR "* -noprofile *". Also, try to add these action-hiding options: https://www.danielbohannon.com/blog-1/2017/3/12/powershell-execution-argument-obfuscation-how-it-can-make-detection-easier

* Let's please remove "hid" or make it stronger. Hid is pretty generic and also has a very special meaning in Windows (Human Interface device) ... and I think it has some meaning in the registry too. To that end, we can pull hid outside of the IN clause, and replace it with a couple of AND/OR conditions "* -w hidden *" OR "* -win hidden *".

patel-bhavin:
looks like this has already been merged in: https://github.com/splunk/security_content/pull/1326/files

P4T12ICK:
@miskoSplunk will close this one because your great feedback was addressed in the other PR

PR #1324, opened 2021-04-12
remove ui and app is not visible

patel-bhavin:
Removed UI for `dist/saaws` package
Updated CI to build a package name called : `DA-ESS_AmazonWebServices_Content`
in app.conf `visible = false`


mvelazc0 (approved):
Looks good

PR #1342, opened 2021-04-21
Password spraying TR-515

mvelazc0:
## New Detections

- Multiple Disabled Users Failing To Authenticate From Host Using Kerberos (Passed)
- Multiple Invalid Users Failing To Authenticate From Host Using Kerberos (Passed)
- Multiple Invalid Users Failing To Authenticate From Host Using NTLM (Passed)
- Multiple Users Attempting To Authenticate Using Explicit Credentials (Passed)
- Multiple Users Failing To Authenticate From Host Using Kerberos (Passed)
- Multiple Users Failing To Authenticate From Host Using NTLM (Passed)
- Multiple Users Failing To Authenticate From Process (Passed)
- Multiple Users Remotely Failing To Authenticate From Host (Passed)

## New Analytic Story

- Active Directory Password Spraying


detections/endpoint/multiple_disabled_users_failing_to_authenticate_from_host_using_kerberos.yml
```diff
@@ -0,0 +1,41 @@
...
+  initial access or elevate privileges. As attackers progress in a breach, mistakes will be made. I certain scenarios, adversaries may execute a password spraying attack against disabled users. 
+  In this logic, the number of unique users is hardcoded to `10` but can/should be customized to fit especific environments.\
+  Event 4768 is generated every time the Key Distribution Center issues a Kerberos Ticket Granting Ticket (TGT). Failure code 0x12 means 
+  `client's credentials have been revoked` (account disabled, expired or locked out).\
+  
+  This detection will trigger on the domain controller against which the offending host performs the password spraying attack.\
+  
+  The analytic's returned fields allow analysts to investigate the event further by providing fields like source ip and attempted user accounts.'
+search: ' `wineventlog_security` EventCode=4768 Result_Code=0x12 | stats dc(Account_Name) AS unique_accounts values(Account_Name) as tried_accounts by Client_Address  | where  unique_accounts > 10 | `multiple_disabled_users_failing_to_authenticate_from_host_using_kerberos_filter` '
```
miskoSplunk: @mvelazc0  This is a nice idea: Same client address, many attempts on blocked accounts.

There is one snag: In companies we have gateway client addresses, i.e., NAT and VPN will affect us here. For example, Splunk's SR office have 2 such gateway IP addresses, SF office has about 5, etc. Thus, all users working from home would start appearing through those IPs - it's similar with cloud.

To that end, could we borrow a time-series component here from @MHaggis (i.e., time buckets) and catch client addresses that generally have less users historically and now have 10 failed ones, and/or client addresses that just  spike in the number of failed accounts suddenly?  

Also, we could add another failure code here (via OR condition) - see if you'd agree (TGT is a precursor for getting a service ticket, so removing a TGT would be the same as revoking the account):
0x14 | KDC_ERR_TGT_REVOKED | TGT has been revoked

Maybe also these - or do them in a separate detection (kind of spoofing/replay detection):
0x43 | KRB_AP_ERR_NO_TGT | No TGT was presented or available
0x22 | KRB_AP_ERR_REPEAT | The request is a replay

detections/endpoint/multiple_invalid_users_failing_to_authenticate_from_host_using_kerberos.yml
```diff
@@ -0,0 +1,41 @@
...
+  initial access or elevate privileges. As attackers progress in a breach, mistakes will be made. In certain scenarios, adversaries may execute a password spraying attack using an invalid list of users. 
+  In this logic, the number of unique users is hardcoded to '10' but can/should be customized to fit especific environments.
+  Event 4768 is generated every time the Key Distribution Center issues a Kerberos Ticket Granting Ticket (TGT). Failure code 0x6 means 
+  'client not found in Kerberos database' (the attempted user is not a valid domain user).\
+
+  This detection will trigger on the domain controller against which the offending host performs the password spraying attack.
+
+  The analytic's returned fields allow analysts to investigate the event further by providing fields like source ip and attempted user accounts.
+search: '`wineventlog_security` EventCode=4768 Result_Code=0x6 | stats dc(Account_Name) AS unique_accounts values(Account_Name) as tried_accounts by Client_Address | where unique_accounts > 10 | `multiple_invalid_users_failing_to_authenticate_from_host_using_kerberos_filter` '
```
miskoSplunk: Same comment ... let's try to overcome the issue of gateway client IPs.

detections/endpoint/multiple_invalid_users_failing_to_authenticate_from_host_using_ntlm.yml
```diff
@@ -0,0 +1,46 @@
...
+  In this logic, the number of unique users is hardcoded to '10' but can/should be customized to fit especific environments.
+  Event 4776 is generated on the computer that is authoritative for the provided credentials. For domain accounts, the domain controller is authoritative.
+  For local accounts, the local computer is authoritative. Error code 0xC0000064 means 'The username you typed does not exis' (the attempted user is a legitimate domain user).\
+
+  This detection will trigger on the domain controller against which the offending host performs the password spraying attack.
+
+  The analytic's returned fields allow analysts to investigate the event further by providing fields like source workstation name and attempted user accounts.
+search: ' `wineventlog_security` EventCode=4776 action=failure 0xC0000064 | stats dc(Logon_Account) AS unique_accounts values(Logon_Account) as tried_accounts by Source_Workstation | where unique_accounts > 10 | `multiple_invalid_users_failing_to_authenticate_from_host_using_ntlm_filter`'
```
miskoSplunk: We probably don't need to guard against gateway IPs, because we'd be getting real device names in Source Workstation fields.

However, let's consider adding these failure codes (in an OR fashion):
0xC0000193 | Account logon with expired account.
0xC0000072 | Account logon to account disabled by administrator.
0xC0000234 | Account logon with account locked.
0xC0000070 | Account logon from unauthorized workstation.

detections/endpoint/multiple_users_failing_to_authenticate_from_host_using_kerberos.yml
```diff
@@ -0,0 +1,45 @@
...
+  This behavior could represent an adversary performing a Password Spraying attack against an Active Directory environment using Kerberos to obtain 
+  initial access or elevate privileges.\ 
+  In this logic, the number of unique users is hardcoded to `10` but can/should be customized to fit especific environments.
+  Event 4771 is generated when the Key Distribution Center fails to issue a Kerberos Ticket Granting Ticket (TGT). Failure code 0x18 means 
+  `wrong passowrd provided` (the attempted user is a legitimate domain user).\
+
+  This detection will trigger on the domain controller against which the offending host performs the password spraying attack.\
+
+  The analytic's returned fields allow analysts to investigate the event further by providing fields like source ip and attempted user accounts.'
+search: '`wineventlog_security` EventCode=4771 Failure_Code=0x18 | stats dc(Account_Name) AS unique_accounts values(Account_Name) as tried_accounts by Client_Address | where unique_accounts > 10 | `multiple_users_failing_to_authenticate_from_host_using_kerberos_filter`'
```
miskoSplunk: This is super sensitive/noise prone both due to gateway IPs (we'd need to again identify that Client Address is not a gateway), as well as due to normal user behaviors on such gateway IPs or multi-user devices - users often mistype their passwords.

I'd suggest removing this detection, if GW/server exclusion ends up being too complicated.

detections/endpoint/multiple_users_failing_to_authenticate_from_host_using_ntlm.yml
```diff
@@ -0,0 +1,46 @@
...
+  initial access or elevate privileges. 
+  In this logic, the number of unique users is hardcoded to '10' but can/should be customized to fit especific environments.
+  Event 4776 is generated on the computer that is authoritative for the provided credentials. For domain accounts, the domain controller is authoritative.
+  For local accounts, the local computer is authoritative. Error code 0xC000006A means 'misspelled or bad password' (the attempted user is a legitimate domain user).\
+
+  This detection will trigger on the domain controller against which the offending host performs the password spraying attack.
+
+  The analytic's returned fields allow analysts to investigate the event further by providing fields like source workstation name and attempted user accounts.
+search: ' `wineventlog_security` EventCode=4776 action=failure 0xC000006A | stats dc(Logon_Account) AS unique_accounts values(Logon_Account) as tried_accounts by Source_Workstation | where unique_accounts > 10 | `multiple_users_failing_to_authenticate_from_host_using_ntlm_filter`'
```
miskoSplunk: Kind of the same comment as for 4771, only that here we only have to tackle/exclude servers ... here we won;t have gateway IPs.

detections/endpoint/multiple_users_failing_to_authenticate_from_one_process.yml
```diff
@@ -0,0 +1,49 @@
+name: Multiple Users Failing To Authenticate From One Process
```
miskoSplunk: Let's reword: The wording "... from one process" seems as if we don't want to tell them what is that "one" process ... sounds semi-informed.

detections/endpoint/multiple_users_failing_to_authenticate_from_one_process.yml
```diff
@@ -0,0 +1,49 @@
...
+ Event 4625 documents each and every failed attempt to logon to the local computer. This event generates on domain controllers, member servers, and workstations. Logon Type 2 
+ describes an iteractive logon attempt.\
+
+ This detection will trigger on the potenfially malicious host, perhaps controlled via a trojan or operated by an insider threat, from where a
+ password spraying attack is being executed. This could be a domain controller as well as a member server orworkstation.
+
+ The analytic's returned fields allow analysts to investigate the event further by providing fields like source process name, source account and attempted user accounts.
+
+search: ' `wineventlog_security` EventCode=4625 Logon_Type=2  !="-" | eval Source_Account = mvindex(Security_ID, 0) | eval Destination_Account = mvindex(Account_Name, 1)  | stats  dc(Destination_Account) AS unique_accounts values(Account_Name) as tried_accounts by Caller_Process_Name, Source_Account, ComputerName  | where  unique_accounts > 10 | `multiple_users_failing_to_authenticate_from_one_process_filter` '
```
miskoSplunk: A couple of notes/suggestions - I'd go with simpler ones first, and then at the end suggest something about this detection as a whole:

* What does '!="-"' mean? Pease note that most events would have "-" field values. Please consider excluding this constraint. for details have a look at example events here: 
https://confluence.splunk.com/pages/viewpage.action?spaceKey=PROD&title=Active+Directory%3A+Identities+and+Locations+of+Logging+for+the+Entities+Reported+in+AD+events

* Why are me making a source user to be a Security_ID? Source user instead is Subject -> Account Name (in TA terminology this would be just Account Name). Now, for most part we won;t care about the source user, and moreover it won't be really reported in 4625s. Instead this field is usually populated with the name of the device where the auth failed (thus, we'd have a dest device here). Further, even for dest device, we have better fields in the event to make extraction - namely ComputerName field. Please see example events in the same confluence page.  

* Next, this part "values(Account_Name) as tried_accounts" makes little sense- let's just please count Destination_Accounts. Those are the accounts which are attempted and failed.

* Now for the main detection: Not sure you'd really want to do just "Logon_Type = 2", reasons being:
1) This will only detect User32 logons, i.e., the logons that access Windows's desktop/logon screen. However, this will exclude all other remote authentications (type 3 and type 8), such as picking up emails, accessing shared resources, remote execution, etc.
2) For Logon Type 2, caller process will always be the same:  C:\Windows\System32\winlogon.exe, so you won't be finding more processess. IN THE END, this may end up just raising noise as many users fail to login to a Windows desktop screen or a server - then this won't be very relevant or even precise. 

* Overall, this detection is not sufficient/insightful if we are doing just "User32 / winlogon.exe" authentication. On the other hand it may be very noise prone if we extend it to logon type 3 and 8, because then we'd be colliding with IT issues - failed servers, expired passwords ... stuff that happens a bunch with Outlook servers.
 
* Let's please meet and brainstorm how to make something better with this detection.

detections/endpoint/multiple_users_failing_to_authenticate_using_explicit_credentials.yml
```diff
@@ -0,0 +1,46 @@
...
+  In this logic, the number of unique users is hardcoded to `10` but can/should be customized to fit especific environments.
+  Event 4648 is generated when a process attempts an account logon by explicitly specifying that account`s credentials. This event generates on domain controllers, member servers, and workstations.
+
+  This detection will trigger on the potenfially malicious host, perhaps controlled via a trojan or operated by an insider threat, from where a
+  password spraying attack is being executed.
+
+  The analytic`s returned fields allow analysts to investigate the event further by providing fields like source account, attempted user accounts and the endpoint were the behavior was identified.'
+
+search: ' `wineventlog_security` EventCode=4648 | eval Source_Account = mvindex(Security_ID, 0) | eval Destination_Account = mvindex(Account_Name, 1) | stats  dc(Destination_Account) AS unique_accounts values(Account_Name) as tried_account by ComputerName | where unique_accounts  > 10 | `multiple_users_failing_to_authenticate_using_explicit_credentials_filter`'
```
miskoSplunk: * Where is a failure component - as it is now,  this detection is not looking at auth failures at all? Event 4648 is mostly a "success" or "info" event. Consequently, please remove this detection.

* Again, Source Account is not Security ID, it would be "mvindex(Account_Name, 0)".

* We have to do pre-filtering because this event is not only about src/dest users: In fact, in src/dest account fields we'd also have devices and service/system accounts. All those should be removed from consideration, i.e., we should add a filter and look only at the cases where mvindex(Account_Name, 0) and mvindex(Account_Name, 1) do not contain dollar "$" at the end, AND are not equal to SYSTEM. See this issue with reports of device and system accounts in event samples in the Confluence page: https://confluence.splunk.com/pages/viewpage.action?spaceKey=PROD&title=Active+Directory%3A+Identities+and+Locations+of+Logging+for+the+Entities+Reported+in+AD+events

detections/endpoint/multiple_users_remotely_failing_to_authenticate_from_host.yml
```diff
@@ -0,0 +1,48 @@
...
+ performing a Password Spraying attack against an Active Directory environment to obtain initial access or elevate privileges. 
+ In this logic, the number of unique users is hardcoded to '10' but can/should be customized to fit especific environments.
+ Event 4625 documents each and every failed attempt to logon to the local computer. This event generates on domain controllers, member servers, and workstations. Logon Type 3 
+ describes an remote authentication attempt.\
+
+ This detection will trigger on the host that is the target of the password spraying attack. This could be a domain controller as well as a member server or workstation.\
+
+ The analytic's returned fields allow analysts to investigate the event further by providing fields like source process name, source account and attempted user accounts.
+
+search: ' `wineventlog_security` EventCode=4625  Logon_Type=3 | eval Destination_Account = mvindex(Account_Name, 1)  | stats dc(Destination_Account) AS unique_accounts values(Destination_Account) as tried_accounts by Source_Network_Address, ComputerName  | where  unique_accounts > 10 | `multiple_users_remotely_failing_to_authenticate_from_host_filter`'
```
miskoSplunk: I'd maybe add Logon Types 8 and 10 here.

miskoSplunk (changes requested):

detections/endpoint/multiple_disabled_users_failing_to_authenticate_from_host_using_kerberos.yml
mvelazc0: thanks @miskoSplunk.

Yes, I wrote these detections knowing that there are some systems like VPN terminators, vulnerability scanners, Citrix farms, exchange servers, etc. that may present this behavior at some point in time. I think the approach there should be to create an allow list and add those known systems to a lookup table. This is a common approach all SOCs implement in some way or fashion. Thoughts ?

detections/endpoint/multiple_users_failing_to_authenticate_from_one_process.yml
mvelazc0: Valid point, changing that to "From Process"

detections/endpoint/multiple_users_failing_to_authenticate_from_one_process.yml
mvelazc0: - Answering first comment: that was a mistake. I just updated it, its: Caller_Process_Name!="-" . On my testing, I noticed that in some scenarios ( For local SYSTEM logon for example), there is no Caller_Process_Name.

detections/endpoint/multiple_users_failing_to_authenticate_using_explicit_credentials.yml
mvelazc0: I updated the query based on the feedback. Also changed the name. This event does not confirm if users are failing to authenticate, just the attempt.

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Disabled Users Failing To Authenticate From Host Using Kerberos :white_check_mark:
- diskUsage: 163840
- runDuration: 0.776s
- scanCount: 22

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Invalid Users Failing To Authenticate From Host Using Kerberos :white_check_mark:
- diskUsage: 163840
- runDuration: 1.031s
- scanCount: 20

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Users Attempting To Authenticate Using Explicit Credentials :x:
- diskUsage: 167936
- runDuration: 0.684s
- scanCount: 20

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Invalid Users Failing To Authenticate From Host Using NTLM :x:
- diskUsage: 204800
- runDuration: 1.072s
- scanCount: 0

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Users Failing To Authenticate From Host Using Kerberos :white_check_mark:
- diskUsage: 163840
- runDuration: 1s
- scanCount: 20

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Users Failing To Authenticate From Host Using NTLM :white_check_mark:
- diskUsage: 241664
- runDuration: 0.811s
- scanCount: 20

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Users Failing To Authenticate From Process :white_check_mark:
- diskUsage: 167936
- runDuration: 0.896s
- scanCount: 20

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Users Remotely Failing To Authenticate From Host :x:
- diskUsage: 163840
- runDuration: 0.71s
- scanCount: 20

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Users Attempting To Authenticate Using Explicit Credentials :white_check_mark:
- diskUsage: 172032
- runDuration: 1.182s
- scanCount: 20

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Invalid Users Failing To Authenticate From Host Using NTLM :white_check_mark:
- diskUsage: 241664
- runDuration: 0.802s
- scanCount: 20

P4T12ICK:
This PR was created by Automated Detection Testing Service :robot:

Tested detections:

Multiple Users Remotely Failing To Authenticate From Host :white_check_mark:
- diskUsage: 167936
- runDuration: 0.657s
- scanCount: 20

detections/endpoint/multiple_disabled_users_failing_to_authenticate_from_host_using_kerberos.yml
mvelazc0: With the new approach, I'm no longer using a numeric threshold but rather standard deviation. 

I dont think it makes sense to add the other Kerberos failure codes. Password Spraying is about obtaining a valid user, hence, result code 0x12 ( bad password ) is the only one that makes sense.

detections/endpoint/multiple_invalid_users_failing_to_authenticate_from_host_using_kerberos.yml
mvelazc0: With the new standard deviation approach, if a gateway client commonly fails to authenticate with user, it will not pass the 3 standard deviations.

Furthermore, I think this detection, just like all others, may require a lookup table or allow list for certain hosts that behave this way. For example, vulnerability scanners.

detections/endpoint/multiple_disabled_users_failing_to_authenticate_from_host_using_kerberos.yml
```diff
@@ -0,0 +1,59 @@
...
+  a Kerberos Ticket Granting Ticket (TGT). Failure code 0x12 means `clients credentials
+  have been revoked` (account disabled, expired or locked out).\
+
+  This detection will trigger on the domain controller against which the offending
+  host performs the password spraying attack.\
+
+  The analytics returned fields allow analysts to investigate the event further by
+  providing fields like source ip and attempted user accounts.'
+search: '`wineventlog_security` EventCode=4768 Account_Name!="*$" Result_Code=0x12
+  | bucket span=2m _time | stats dc(Account_Name) AS unique_accounts values(Account_Name)
```
miskoSplunk, line 27: Mauricio, would it hurt if we moved this to 30 minutes (also for other detections)? Reason being, although some events are indeed frequent (4768s are one of the most frequent), they may not be this frequent for many accounts to start failing in 2 minutes.

Try this is DEV-RZ ... most devices would have just one 2-minute slot with 4768s in 30 minutes of observation,  and inside those slots there will be one or 2 events.
`sourcetype="WinEventLog" OR sourcetype="wineventlog" index="*" EventCode=4768 Account_Name!="*$" | bucket span=2m _time | stats count by Client_Address, _time | eventstats count as numSlots by Client_Address | where numSlots > 0`

detections/endpoint/multiple_users_attempting_to_authenticate_using_explicit_credentials.yml
```diff
@@ -0,0 +1,66 @@
...
+  and workstations.\
+
+  This detection will trigger on the potenfially malicious host, perhaps controlled
+  via a trojan or operated by an insider threat, from where a password spraying attack
+  is being executed.\
+
+  The analytic`s returned fields allow analysts to investigate the event further by
+  providing fields like source account, attempted user accounts and the endpoint were
+  the behavior was identified.'
+search: ' `wineventlog_security` EventCode=4648 | bucket span=2m _time | eval Source_Account
```
miskoSplunk: Mauricio, this one would definitely need slots longer than 2 minutes ... see dev-rz, for SV5-ITO-OKTA001.splunkcorp.com which is a global auth server, and source account srv-ito-okta (in one slot it authenticates 9 accounts, then 21, then 12 .... until avg and stdev stabilize (if at all) it would throw many unnecessary detections.

`sourcetype="WinEventLog" OR sourcetype="wineventlog" index="*"EventCode=4648 | bucket span=2m _time | eval Source_Account
  = mvindex(Account_Name, 0) | eval Destination_Account = mvindex(Account_Name, 1)
  | search Source_Account != "*$" Source_Account !="-" Destination_Account !="*$"
  | stats dc(Destination_Account) AS unique_accounts values(Destination_Account) as
  tried_account by _time, ComputerName, Source_Account | eventstats count as numSlots by ComputerName, Source_Account | where numSlots > 1`

detections/endpoint/multiple_users_failing_to_authenticate_from_process.yml
```diff
@@ -0,0 +1,68 @@
...
+
+  This detection will trigger on the potenfially malicious host, perhaps controlled
+  via a trojan or operated by an insider threat, from where a password spraying attack
+  is being executed. This could be a domain controller as well as a member server
+  orworkstation.\
+
+  The analytic''s returned fields allow analysts to investigate the event further
+  by providing fields like source process name, source account and attempted user
+  accounts.'
+search: ' `wineventlog_security` EventCode=4625 Logon_Type=2 Caller_Process_Name!="-"
```
miskoSplunk, line 28: Maybe we can do specific error codes like for other events ... some interesting ones could be
Event ID 4625 status codes:

0xC0000064 User logon with misspelled or bad user account
0xC000006A User logon with misspelled or bad password
0XC000006D This is either due to a bad username or authentication information
0XC000006E Unknown user name or bad password.

miskoSplunk (changes requested):

detections/endpoint/multiple_disabled_users_failing_to_authenticate_from_host_using_kerberos.yml
mvelazc0: Misko, on my initial testing once I was working with the STDEV approach, I was using a larger bucket, I think 15 minutes. I found several false positives with that. I tested many different combinations on Dev RZ and uba-perf, the best results came with 2 minutes buckets.

On most of the scenarios I have encountered password spraying attacks, they are usually performed in a small period of time as attackers target a large number of users to increase the chances of finding valid credentials. Thats why I think it makes sense to measure what is "normal" in a small bucket of time.

Furthermore, we plan to configure this search so that it runs every hour. So, there will be 30 two-minute slots of time to identify observations and calculate the standard deviation. If we change this bucket to 30 minutes. there would only be 2 slots of time and this would affect the standard deviation calculation.

detections/endpoint/multiple_users_attempting_to_authenticate_using_explicit_credentials.yml
mvelazc0: Misko, I think this is expected. Enterprise environments will always have multi-user systems where several users will be authenticating (or failing to authenticate) from one source. These include but are not limited to VPN terminators, Identity providers, Exchange servers, Web applications like an Intranet, Citrix farms, etc.  

On the false positive documentation for this detection, I clearly describe that this type of systems will have to be identified and added to an allow list (https://github.com/splunk/security_content/blob/password_spraying_TR-515/detections/endpoint/multiple_users_attempting_to_authenticate_using_explicit_credentials.yml#L35 ). This is not anything new for SOCs, its normal to have exceptions for certain detections. In the case of Dev RZ, once you add these servers to an allow list, there is no other systems showing this behavior. 

Finally, since we plan to schedule this query to run every hour, it should only take the first time to allow analysts to identify these systems and whitelist them.

detections/endpoint/multiple_users_failing_to_authenticate_from_process.yml
mvelazc0: This is a good idea. Perhaps we can use this initial search to create two or three more specific ones with the different error codes.

However, I will leave that for a next iteration of this detection. We do want to push an initial version this week.

josehelps (approved):
looks good to me

josehelps (approved):

josehelps (approved):

miskoSplunk (approved):

PR #1353, opened 2021-04-28
We're Off To See The Haag

MHaggis:
Modified all SSA detections `body` from `tbd` to include fields.


josehelps (approved):
looks good to me, removed some dest_user_id since this is already under anomaly output.

PR #1354, opened 2021-04-29
Feature/generate reorg

nbtnc1:
The general idea behind this PR is to split up the logic into different functions. We'd like to reuse the parts of this code that import and compute the data structure, but then use the Splunk API to upload, rather than generate the conf files.

- generate_* functions now only do Jinja outputting - no compute logic.
- Compute logic has been moved to prepare_* functions.
- Removed replace_vars_in_deployment as I believe custom_jinja2_enrichment_filter should be used instead?
- Added import_objects to import all the yaml data in an objects dictionary.

Feel free to modify etc. I've tested by comparing the .conf file outputs before and after the PR but please double check.


P4T12ICK:
Thank you very much. We will review it.

josehelps:
@nbtnc1 This is wonderful! We will definitely take the changes in once we have tested them thoroughly. Give us a few days to do this.

P4T12ICK (approved):
Looks good to me. Thank you very much @nbtnc1

PR #1380, opened 2021-05-07
Detection prefixes and import config

nbtnc1:
Another change that is useful for us and thought I would push back to you. Feel free to use if you want.

All of the current searches have "ESCU" as a prefix and we wanted to add our own detections with a different prefix.
Added a prefix variable to all objects and created a config.yml file to allow imports from other locations without changing the code.


josehelps:
Closing for now since this PR has gone stale. Feel free to re-open when it is ready to merge.

PR #1423, opened 2021-05-28
Clear out old CI pipelines

peter-cg:
We maxed out on DSP pipelines in staging so I added code to get rid of old ones (>24h) which were lingering for whatever reason.

And some small logging visibility/debug changes.


P4T12ICK:
Thank you @peter-cg. That is a good idea. In theory, we would not need that code but sometimes the clean up doesn't work for whatever reason and therefore this is a great idea. I will review it.

P4T12ICK (approved):
Looks good to me

PR #1447, opened 2021-06-16
TR-723 fixing bodies error

josehelps:
mispelling of the variable name.


peter-cg (approved):
I verified that this one compiles and the old one doesn't.

PR #1448, opened 2021-06-16
Update ssa___detect_kerberoasting.yml

josehelps:
was missing write into ssa as well as extracting the device ids.


jzsplunk (approved):
LGTM.  Readability wise you break a long line down into individual commands sometimes makes it a little easier to catch gotchas.

PR #1444, opened 2021-06-16
ssa

tccontre:
## Endpoint to DSP SPL2 Detection (Conversion and test)

1. Wevtutil Usage To Disable Logs
2. WevtUtil Usage To Clear Logs
3. Deny Permission using Cacls Utility
4. Grant Permission Using Cacls Utility
5. Modify ACLs Permission Of Files Or Folders

Wevtutil Usage To Disable Logs
dis_evt

WevtUtil Usage To Clear Logs
clr_evt

Deny Permission using Cacls Utility
deny_ssa

Grant Permission Using Cacls Utility
grant_ssa

Modify ACLs Permission Of Files Or Folders
acl_mod

PR #1462, opened 2021-06-24
[PR-1444] fix branch reversion.

jzsplunk:
@tccontre @patel-bhavin @peter-cg FYI the branch for https://github.com/splunk/security_content/pull/1444/files with the proper ssa_ prefix to get into CI


tccontre:
it looks like the test passed on this PR. :) @peter-cg @patel-bhavin @jzsplunk

peter-cg (commented):
Since you're doing string equality checks it might be good to do something like
`process_name=lower(ucast(map_get(input_event, "process_name"), "string", null))`
unless process_name is known to be lower-case already somehow.

tccontre:
thanks @peter-cg . as we discussed offline Ill do this minor improvement after the merging of the detection.

patel-bhavin (approved):

PR #1464, opened 2021-06-25
doc_gen.py issues with pyattck

josehelps:
removing pyattck and changing `doc_gen.py` to instead use the STIX CTI library like we do with generate-actor-map.py and `generate-coverage-map.py`. This should resolve CI failing issues.


tccontre (approved):
LGTM

PR #1470, opened 2021-07-01
Apger first change to content ./endpoint/winword_spawning_powershell.yml

apger:
I don't use git that often outside of my own project but here is my first attempt at contributing to the greater Splunk good.  please advise and I get to work if attempt is inline with your process.


detections/endpoint/winword_spawning_powershell.yml
```diff
@@ -52,3 +52,24 @@ tags:
...
   - parent_process_id
   security_domain: endpoint
+  impact: 5000
+  confidence: 1
+  # (impact * confidence)/100
+  risk_score: 50
+  context:
+  - Source:Endpoint
+  - Stage:Execution
+  message: $parent_process_name$ on $dest$ by $user$ launched the following powershell process: $process_name$ which is very common in phishing attacks.
```
patel-bhavin: Hey @apger : looks like our validation is failing on the `value`of the message, can you change `process:` to `process-`

josehelps: alternatively tie that around quotes so it gets treated as a string :smile:

apger: Fixed with Bhavin

apger:
Changes made - new pull request coming.

PR #1495, opened 2021-07-07
all 5 detections should be perfect

apger:
Fixed the multi stage issue per team feedback.  5 rules modified.  Passes validation.


apger:
I think this is related to the multi-stage scenario for a single detection.  I changed it from multiple comma separated values in a single "stage=" to multiple comma separate "stage=" assignments.  I will change it and retest.

apger:
Started a new branch from scratch.  I think this is a build problem; not a content problem.

PR #1506, opened 2021-07-12
TR-772 - Rba support

mvelazc0:


detections/endpoint/malicious_powershell_process___encoded_command.yml
```diff
@@ -56,3 +56,18 @@ tags:
   - Processes.dest
   - Processes.process_id
   security_domain: endpoint
+  impact: 70
+  confidence: 50
+  risk_score: 35
+  context:
+  - Source: Endpoint
```
josehelps: minor thing, all of these under context are meant to be simple string eg: `Source:Endpoint` no space in between.

detections/endpoint/suspicious_driver_loaded_path.yml
```diff
@@ -50,3 +50,19 @@ tags:
...
+  - Source: Endpoint
+  - Stage: Execution, Defense Evasion
+  message: Suspicious driver $ImageLoaded$ on $Computer$
+  observable:
+  - name: Computer
+    type: Endpoint
+    role:
+    - Victim
+  - name: ImageLoaded
+  type: File Name
```
josehelps: this seems spaced incorrectly

josehelps (changes requested):
two minor things that need updating, otherwise it looks great.

detections/endpoint/malicious_powershell_process___encoded_command.yml
mvelazc0: fixed these across all detections

detections/endpoint/suspicious_driver_loaded_path.yml
mvelazc0: this was also fixed

josehelps (dismissed):

PR #1530, opened 2021-07-19
add analytic types to detections

P4T12ICK:
add analytic types to detections


patel-bhavin:
hey @P4T12ICK : i tested this on a fresh range and the PR looks solid!

patel-bhavin (approved):

PR #1536, opened 2021-07-21
Filter PTR requests to optimize deviation calculation and remove them from result.

sec-researcher:
As I know PTR requests can not be used for data exfiltration so having them in search result is just a false positive.
Also they have effect on deviation calculation and lead to a lot of false positive in result, specially when PTR requests in the environment is about 20% or more. So I add this filter to the search 'where NOT DNS.message_type IN("Pointer","PTR")'


patel-bhavin:
Thank you for the PR. We ran this SPL update through our detection testing service and it passed!

patel-bhavin (approved):

PR #1539, opened 2021-07-22
Update Condition on detect_processes_used_for_system_network_configuration

BlackB0lt:
Except user is NULL or unknown. It will decrease false detection


patel-bhavin:
Hello, 
Thank you for your contribution, since there are SPL changes in your: we would like to get them tested via our detection testing service. I create a duplicate PR https://github.com/splunk/security_content/pull/1545

PR #1546, opened 2021-07-28
TR-770 added event_id to ssa detections

josehelps:
this will allow alerts to be traced back to their original events in SSA.


peter-cg (changes requested):
These don't currently compile. I think the `event_id` field needs to be extracted from the `input_event`.

peter-cg:
Superseded by this PR: https://github.com/splunk/security_content/pull/1550

PR #1578, opened 2021-08-21
Filter blocked traffic and add some recommendation to "how_to_implement" section

sec-researcher:
Filter blocked traffic because it made a lot of false positive specially when something like port scan happen, also add description to "how_to_implement" section to limit result just to organization private and public IP range for more accurate result.


MHaggis:
Thank you @sec-researcher. We will get this reviewed and tested.

MHaggis (dismissed):

PR #1594, opened 2021-08-26
Content ctl updates

patel-bhavin:
Updates types in detections 
test file path errors that we encountered @mvelazc0 
removed baseline.j2 since we dont have a baseline as a folder
removed type from story file


MHaggis (dismissed):

josehelps (approved):
LGTM, added risk datamodel

PR #1609, opened 2021-08-30
Gfiledetections

rosplk:
pr for devsecops and gsuite phishing


detections/cloud/detect_gdrive_doc_scanning.yml
```diff
@@ -0,0 +1,49 @@
+name: Detect Gdrive Doc Scanning
```
patel-bhavin, line 1: We cannot write a proper detection for this since after i look at RZ data: there are 100s apps/clients (at Splunk) that scans(view/download) for gdocs. 

With a simple detection like this will create way too many FPs

detections/cloud/gdrive_suspicious_file_sharing.yml
```diff
@@ -0,0 +1,51 @@
...
+id: b842f557-c7b3-4b92-aa34-8243217a6685
+version: 1
+date: '2021-08-19'
+author: Rod Soto, Splunk
+type: Anomaly
+datamodel:
+- N/A
+description: This search can help the detection of compromised accounts or internal users sending malicious documents via GSuite file sharing.
+search: 'sourcetype="gsuite:drive:json" name=change_user_access parameters.target_user > 100 | spath "parameters.owner" | search "parameters.owner"="*" | stats  count by src_ip parameters.owner parameters.target_user parameters.doc_type parameters.doc_title | `security_content_ctime(firstTime)`| `security_content_ctime(lastTime)`
+  | `gdrive_suspicious_file_sharing_filter`'
```
tccontre, line 11: as we have discussed yesterday with Bhavin. just use distinct stats if you want to check the number of events in given time, focus first on small scenario like suspicious file sharing from outside email going in then try to reduce FP. also this checking "parameters.target_user > 100" must be place after the stats command where you already have your statistics of the field you want to check. :) thanks man!

detections/cloud/gsuite_suspicious_calendar_invite.yml
```diff
@@ -0,0 +1,54 @@
...
+id: b842f557-c7b3-4b92-aa34-8243217a6685
+version: 1
+date: '2021-08-19'
+author: Rod Soto, Splunk
+type: Anomaly
+datamodel:
+- N/A
+description: This search can help the detection of compromised accounts or internal users sending suspcious calendar invites via GSuite calendar. These
+invites may contain malicious links or attachments.
+search: 'sourcetype="gsuite:calendar:json" email=* parameters.event_title!=null parameters.target_calendar_id > 100 | stats  count by email ip_address source name parameters.api_kind parameters.event_title parameters.organizer_calendar_id parameters.target_calendar_id parameters.event_response_status   | convert  timeformat="%Y-%m-%dT%H:%M:%S" ctime(firstTime)
```
tccontre, line 11: same as the comment above. :)

josehelps:
closing PR for now since we are not sure about the usage of these searches just yet.

PR #1631, opened 2021-09-03
Adding a Pth/PtT detection from a vantage point of domain controllers

miskoSplunk:


miskoSplunk:
closing to rename the banch, so it can be tested in SSA

PR #1678, opened 2021-09-23
Exlude windows DNS query for finding AD services

sec-researcher:
Windows generate a lot of DNS query for finding AD services these queries have a long TLD and made a lot of false positives, but on the Internet currently the longest TLD has 24 chars’ length, so by adding this line queries with TLD longer than 24 chars will be excluded.


sec-researcher:
Something went wrong on detection testing:
 Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers

patel-bhavin:
as of now, automated detection testing workflow is only available to users from Splunk organization. I have created a duplicate PR : https://github.com/splunk/security_content/pull/1703 to run it via our pipeline.

Like always, thanks so much for helping improving our detections

sec-researcher:
So since now I can't create PR?

PR #1733, opened 2021-10-22
Commit of risk notable playbooks

kelby-shelton:
- Copied py and json from community
- Added screenshots for each playbook
- Created yml for each playbook


ljstella:
Cleaned up & merged as part of #1793

PR #1743, opened 2021-10-27
fixed issue with ATTACK table not displaying correctly

josehelps:
minor bug with attack table on docs.


josehelps:
fixes this table 
image

ljstella (approved):
LGTM

PR #1773, opened 2021-11-01
Feature/pex 33 api lambda scalability issue

dleung-splunk:
Yaml2Json is a logical port of security-content-api > chalicelib > data_mapper.py
This module processes YAML files directly within the repo and outputs consolidated JSON.

I tried adding this to the build (build-and-validate.yml) however during the run it could not locate the file.
```
Run source venv/bin/activate
python3: can't open file '/home/runner/work/security_content/security_content/bin/yaml_to_json.py': [Errno 2] No such file or directory
Error: Process completed with exit code 2.
```

My guess is this file needs to be merged with develop before it can be used in the build  -- I'm guessing the build runs on develop.


bin/yaml_to_json.py
```diff
@@ -0,0 +1,275 @@
+import yaml, json
+from chalice import BadRequestError
```
josehelps: remove this requirement since, it won't run with the requirements.txt currently on the project since it does not include it.

bin/yaml_to_json.py
```diff
@@ -0,0 +1,275 @@
...
+                if not ('fields_list' in lookup_obj):
+                    csv_file_name = lookup_obj['filename']
+                    lookup_obj['csv_file_url'] = 'https://security-content.s3-us-west-2.amazonaws.com/lookups/' + csv_file_name
+
+                lookup_objects.append(lookup_obj)
+
+        return lookup_objects
+
+
+if __name__ == "__main__":
```
josehelps, line 258: I would include argument parsing here that:
* takes in the project repo location (you can have it default to `.`)
* takes in an output path you can have it default to`yaml_to_json`
see the following script as an example: https://github.com/splunk/security_content/blob/develop/bin/doc_gen.py#L371-L378

josehelps (changes requested):
I would also refactor this code to be part of the `generate.py` https://github.com/splunk/security_content/blob/develop/bin/generate.py, alot of this logic already exists there. This can be exposed as another "product" via contentctl.py https://github.com/splunk/security_content/blob/develop/contentctl.py#L114 .. for example `--product API` and save the jsons to `dist/api` directory.

bin/yaml_to_json.py
dleung-splunk: Fixed.

bin/yaml_to_json.py
dleung-splunk: > I would also refactor this code to be part of the `generate.py` https://github.com/splunk/security_content/blob/develop/bin/generate.py, alot of this logic already exists there. This can be exposed as another "product" via contentctl.py https://github.com/splunk/security_content/blob/develop/contentctl.py#L114 .. for example `--product API` and save the jsons to `dist/api` directory.

@d1vious,
As per your suggestion, this is now callable from `contentctl.py` as shown:

    `python3 contentctl.py --path . --verbose generate --product API --output dist/api`

I refactored `yaml_to_json.py` to work from within `generate.py`.  I didn't try to copy and paste the contents over into `generate.py` as there was not any good overlap of existing functionalities.  Instead, I simply modified and imported `yaml_to_json.py` as a module called from within `generate.py` and indirectly callable from `contentctl.py`.

Moreover, I also updated `yaml_to_json.py` to accept the params for repo location and output path for JSONs.  As per your suggestion, we're calling this new functionality.

For lack of imagination, the API JSONs will be uploaded to a new folder called `json` on [s3/security-content](https://s3.console.aws.amazon.com/s3/buckets/security-content?region=us-west-2&prefix=json/&showversions=false), during production builds:
`aws s3 cp dist/api s3://security-content/json --recursive --exclude "*" --include "*.json"`

patel-bhavin (approved):

patel-bhavin (approved):

PR #1783, opened 2021-11-05
Fixing lowercasing of entitiy ids, which must not be done for SSA.

miskoSplunk:
Fix for lowercasing of entity IDs which should not be done in SSA


ljstella (approved):
LGTM

PR #1796, opened 2021-11-10
Update aws_iam_accessdenied_discovery_events.yml

infosecB:
Remove redundant "values(userIdentity.arn)" from stats command. The field is included in "by" section of stats command.


patel-bhavin:
Thanks for creating this PR, since this had an SPL update: we would need to run it through our detection testing pipeline. Created a duplicate PR for that https://github.com/splunk/security_content/pull/1832

PR #1853, opened 2021-12-01
Gone in 60 LOLBAS

josehelps:
major overhaul to the detection, change logic to use distinc count, added more objects to the body that make the alert useful, as well as renaming, and overhauling description


trogdorsey (changes requested):
The streaming version of stats does not support estdc.

Message references cmd_line, which isn't referenced in the detection itself. And contains a typo (iin)

PR #1864, opened 2021-12-06
Convert Detection Testing from Batch to GH-Actions Docker

pyth0n1c:
This is a significant change to our testing pipeline.  It paves the
way to eliminate AWS Batch, which requires manual approval
and runs one at a time, from the CI process and replaces it 
with Splunk Docker containers running in GitHub Actions.
This lets us run more tests and faster, especially for large
sets of detections.  This will run in parallel with the AWS
Batch detections until we have validated the performance
of the new detection testing system.


pyth0n1c:
This branch has been deprecated by the DockerDetectionUpdatesAndFixes branch, which has a huge number of updates and fixes.

PR #1874, opened 2021-12-08
BA detections

P4T12ICK:
New detections:
- Excessive Number of Office Files Copied
- High File Deletion Frequency


josehelps (approved):
LGTM

PR #1875, opened 2021-12-08
Anomalous usage of account credentials

ljstella:
Tied to BA Insider Threat work. This detection is under experimental for two reasons:

- It relies on estdc() instead of dc() which is an undocumented feature of SPS
- the lack of values() non-numeric aggregation feature at this time, which destroys any context related to this anomaly.


josehelps:
@ljstella given the two limitations above can we:
1. move it to stable by giving a test and a data set
2. add to the body a field called "drilldown" with a string of a search that would allow the user to find the context before the stats command?

josehelps (changes requested):
lets add a test file for this and move it to stable

ljstella (commented):
Dataset & test generated and added, moved back to experimental per conversation w/ @d1vious

josehelps (approved):

patel-bhavin (approved):

PR #1903, opened 2021-12-22
WIP: Enriched detections with supprted_tas and tas_with_data in detection files

truptilangalia-crest:
Following changes have been done as part of this MR : 

- Added script to enrich detections
- added a detection-ta-mapping.yml 
- Enriched detections with supported_tas and tas_with_data lists


josehelps:
Two comments, here:
1. looks like the way we are calculating supported_tas is not taking into account context of the detection, for example here we recommend supported_ta for Splunk_TA_nix and it is a windows only detection:   https://github.com/splunk/security_content/pull/1903/files#diff-44974be75facadaedddede27dd346df6c0fdf3dce5d91c809472d193ce96ee9a
2. we should change the names of the field, supported_tas should have the "tested" TAs, and maybe rename what we call supported to recommended_tas or similar, but I would not call supported things we have not tested.

truptilangalia-crest:
> Two comments, here:
> 
> 1. looks like the way we are calculating supported_tas is not taking into account context of the detection, for example here we recommend supported_ta for Splunk_TA_nix and it is a windows only detection:   https://github.com/splunk/security_content/pull/1903/files#diff-44974be75facadaedddede27dd346df6c0fdf3dce5d91c809472d193ce96ee9a
> 2. we should change the names of the field, supported_tas should have the "tested" TAs, and maybe rename what we call supported to recommended_tas or similar, but I would not call supported things we have not tested.

1. Yes we are not taking context of the detection into account while calculating supported_tas. You can find the whole logic that we are using for calculating supported_tas here inside "Steps to enrich detection" : https://docs.google.com/document/d/1t0vSKhcsNgw_t36t10IHYjRNLOq4lEk5n9obxOWyYxU/edit?usp=sharing
2. We have updated the supported_tas to recommended_tas

P4T12ICK:
I just did a review. I totally agree with Jose that the tas_with_cim_mapping doesn't take the context of a detection into account. And this leads to a false impression to our users that one of tas_with_cim_mapping could be used instead, which is wrong even when theoretically the TA parses fields with the same name. 
Most of our detections only works for the specific data source (e.g. Sysmon) used for development even though a CIM datamodel is used in the search.
Additionally, I'm not sure if the values are correct. Most of our endpoint detections uses Sysmon as the data source. I can't see this value in supported_tas.

P4T12ICK:
The other problem what I see here is that even when the data would be correct. We spent a lot of time to keep our yml short and readable in order to make content authoring as easy as possible. Adding 60 additional lines for some detections is not okay e.g. powershell_execute_com_object.yml .

truptilangalia-crest:
> The other problem what I see here is that even when the data would be correct. We spent a lot of time to keep our yml short and readable in order to make content authoring as easy as possible. Adding 60 additional lines for some detections is not okay e.g. powershell_execute_com_object.yml .

It was becasue that detection was having only "_time" as required field so all tas were mapping in that case. But now we have updated the logic, so that if there is no other required fields other than "_time" than it will skip enriching that detection.

josehelps:
@truptilangalia-crest the PR looks good to me, thank you so much for the hard work! :heavy_check_mark:

josehelps (dismissed):

P4T12ICK:
@truptilangalia-crest can you do one minor change? Can we remove the cim_version field for now? It think it doesn't make a lot of sense in the moment.

truptilangalia-crest:
> @truptilangalia-crest can you do one minor change? Can we remove the cim_version field for now? It think it doesn't make a lot of sense in the moment.

Yes, I have removed the cim_version for now.

josehelps (approved):

P4T12ICK (approved):

PR #1919, opened 2022-01-05
Playbook migration for remcos

ljstella:
In preparation for upcoming content release, migrating 3 playbooks to `security_content`


playbooks/block_indicators.yml
```diff
@@ -0,0 +1,23 @@
...
+references: []
+app_list:
+- "Palo Alto Networks Firewall"
+- "CarbonBlack Response"
+- "OpenDNS Umbrella"
+tags:
+  platform_tags:
+  - Response
+  playbook_fields:
+  - DestinationDnsDomain
```
philroyer-phantom: I think these three fields should use the same capitalization as SOAR uses for all CEF fields, so they should be destinationDnsDomain, destinationAddress, and fileHash

playbooks/email_notification_for_malware.yml
```diff
@@ -0,0 +1,22 @@
...
+app_list:
+- "VirusTotal"
+- "WildFire"
+- "CarbonBlack Response"
+- "SMTP"
+tags:
+  platform_tags:
+  - Response
+  playbook_fields:
+  - FileHash
```
philroyer-phantom: Should be fileHash because SOAR is case-sensitive

philroyer-phantom: Should also have the vaultId field which is used for "detonate file"

playbooks/malware_hunt_and_contain.yml
```diff
@@ -0,0 +1,22 @@
...
+app_list:
+- "LDAP"
+- "ServiceNow"
+- "CarbonBlack Response"
+- "VirusTotal"
+tags:
+  platform_tags:
+  - Response
+  playbook_fields:
+  - FileHash
```
philroyer-phantom: Should be filehash

philroyer-phantom (changes requested):
Just some small changes around playbook_fields then it should be good

playbooks/block_indicators.yml
ljstella: Fixed

playbooks/email_notification_for_malware.yml
ljstella: Fixed

playbooks/malware_hunt_and_contain.yml
ljstella: Fixed

philroyer-phantom (approved):
👍

josehelps (approved):
LGTM

PR #1947, opened 2022-01-20
BA add body field with Finding Report Object

P4T12ICK:
BA add body field with Finding Report Object


josehelps (dismissed):
LGTM

subhakar:
@d1vious / @P4T12ICK - Few things to incorporate in the changes made to the detections

1. There are a bunch of changes made to the findingReport schema that need to be reflected here - http://event-schema.splunk.com/classes/splunk/finding_report
2. As we make changes to the detections to generate finidingReport events we want to use the sink - `write_ssa_finding_events()`. This will differentiate the sinks to write the events appropriately
3. **This is on the BA sink expansion -** With the current changes, we are still going to generate anomaly events. The only difference is that the body now contains the findingReport object. Instead we would like to simply return the findingReport object. This can be done by replacing `serialize_json()` with `to_bytes(to_json(map))` where map is the findingReport object

Please let me know if you have any questions and we can have a quick sync up.

josehelps:
stale closing for now

PR #1972, opened 2022-01-31
Update o365_bypass_mfa_via_trusted_ip.yml

ionsor:
Modified the "Signature" field to "Operation" to match the search with the test data. And removed other fields from the stats count which don't exist in the test data


patel-bhavin:
Duplicate created for testing: https://github.com/splunk/security_content/pull/1993. 
Thanks for the update

PR #1973, opened 2022-01-31
Update o365_disable_mfa.yml

ionsor:
Modified the search to match the test data. "signature" field doesn't exist in test data, so I modified it to "Operation" field to match.


patel-bhavin:
Duplicate created for testing: https://github.com/splunk/security_content/pull/1993. 
Thanks for the update

PR #1976, opened 2022-02-01
fixes TR-1309

josehelps:
I think this resolves a weird UI bug in ES for risk events.


patel-bhavin:
updated Jinja and tested on Fonder! works nnow

patel-bhavin (approved):

PR #1985, opened 2022-02-02
Fix detection BA Hiding Files And Directories With Attrib exe

P4T12ICK:
Fix detection BA Hiding Files And Directories With Attrib exe


P4T12ICK:
Screenshot 2022-02-02 at 10 31 14

tccontre (approved):
LGTM

PR #1994, opened 2022-02-04
Saving Private LookupMacros

josehelps:
# Adds Lookups and Macros to research.splunk.com

1. adds both objects to each detection page
2. adds a link to the type for each detection
3. adds an explanation for empty macros
4. moves around a bit how we display info in the detection page.


ljstella (approved):
LGTM

PR #2036, opened 2022-02-18
Docgen improvements

pyth0n1c:
When 'https://cve.circl.lu' is running slowly, it can cause 
the doc_gen.py script, which is part of our CI/CD workflow,
to either run EXTREMELY slowly or seemingly freeze.  This
is an API endpoint that we use to enrich our CVEs.  
We had added two features:
1.  Memoize/cache the results from calls to this API.  This
makes subsequent calls for a CVE we already called return
instantly.
2.  Add a timeout to call to this endpoint. The timeout is 
currently 10 seconds, but this may need to be adjusted
in the future during testing.

### DIFF
--- a/bin/doc_gen.py
+++ b/bin/doc_gen.py
@@ -1,3 +1,4 @@
+from threading import Thread
 from functools import cache
 import glob
 import yaml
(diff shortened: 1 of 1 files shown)


ljstella (approved):
LGTM

PR #2064, opened 2022-03-03
remove CI step

patel-bhavin:


pyth0n1c (approved):
Removed for debate on how to correctly pass token

PR #2085, opened 2022-03-16
Search name check- contentctl

patel-bhavin:
updated contentctl validate code for detection and baselines to check for character length for ESCU searches
Truncated all searches to have names less that 67 chars for ESCU detections only

### DIFF
--- a/bin/contentctl_project/contentctl_core/domain/entities/detection.py
+++ b/bin/contentctl_project/contentctl_core/domain/entities/detection.py
@@ -52,11 +52,11 @@ class Detection(BaseModel, SecurityContentObject):
     source: str = None
 
 
-    @validator('name')
-    def name_max_length(cls, v):
-        if len(v) > 67:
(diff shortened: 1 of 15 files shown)


bin/contentctl_project/contentctl_core/domain/entities/detection.py
```diff
@@ -112,6 +112,14 @@ def search_validation(cls, values):
                     raise ValueError('Use source macro instead of eventtype, sourcetype, source or index in detection: ' + values["name"])
         return values
 
+    @root_validator
+    def name_max_length(cls, values):
```
patel-bhavin, line 116: @P4T12ICK : Hey, i added the name_max_length function under root validator and only checks for ESCU search name length. I tested this and it works correctly, do you think this is how you'd do this check?

P4T12ICK (approved):
looks good. good work

PR #2089, opened 2022-03-17
Fixed issue with github action workflow for enrich-detections

truptilangalia-crest:
Changed the name for github token used in github action and uncommented the code as previous variable name was not allowed to be stored in action secrets and hence it was causing error.


patel-bhavin:
as discussed, we are no longer including this as a workflow step but instead as a scheduled job that create a PR automatically. Closing this PR for now

PR #2103, opened 2022-03-22
cleaning up old CI jobs

josehelps:
removes:
- [x] old aws batch CI/CD job
- [x] old BA CI/CD jobs
- [x] renames detection-testing CI to docker-detection-tester


pyth0n1c (approved):
Everything looked good, but needed to update the
detection-testing.yml workflow file. That's been done 
as well. I expect that detection testing will fail since 
we always pull and run that from develop, so we will
merge before those tests complete and re-evaluate the
tests that run after the merge.

PR #2108, opened 2022-03-24
TR-1714 add supported tas to research site

josehelps:
this adds a new enrichement to contentctl that takes the `supported_tas` tag and finds the splunkbase app name and url. This then gets rendered on the research site. Also updates the jinja2 templates to properly render things under a new **Datasource** field. 

image


P4T12ICK (approved):
Looks good to me

PR #2111, opened 2022-03-24
content_ctl generate : api 

patel-bhavin:
- bug fixes while creating json files for API 
-removes `file_path` from the json files
updated all test json files


patel-bhavin:
@P4T12ICK : this fixes the api object creation that we spoke about. please have a look when you're back!

PR #2110, opened 2022-03-24
fixing lyonell

josehelps:
# PR Template for new Detections

For Authors:
- [ ] Make sure that CI/CD [detection-testing and build-and-validate](https://github.com/splunk/security_content/actions) jobs passed ✔️. 

For Reviewers:
- [ ] Verify CI/CD jobs have passed without errors.
- [ ] Validate SPL logic.
- [ ] Validate tags, description, and how to implement.
- [ ] Validate name matches `<platform>_<mitre att&ck technique>_<short description>`
- [ ] Verify references match analytic.
- [ ] Is there an Atomic Test?

PR #2125, opened 2022-03-28
Adding CIS and Nist tags to BA

patel-bhavin:
fix for- https://splunk.atlassian.net/browse/TR-1801
and includes this change as well : https://github.com/splunk/security_content/pull/2110


ljstella (approved):
LGTM

PR #2190, opened 2022-05-10
kvstore bugs

patel-bhavin:
Encountered a bug in generate while debugging :https://splunk.atlassian.net/browse/TR-1279

This PR adds kvstore related definitions into transforms correctly


pyth0n1c (approved):
Detection is failing due to known issue with baselines.
However, this is a known issue that is being actively worked.
Approving updates to detection documentation.

Approving updates to jinja2 template.

PR #2207, opened 2022-05-20
adding insider threat, fixing minor display bug

josehelps:
just fixing some research.splunk.com bugs, mainly link to splunk product page, adding insider threat and new actor map.


josehelps:
using admin powers :smile:

PR #2218, opened 2022-05-26
updating mitre ids and other metadata

patel-bhavin:
Updates Mitre IDs
Updates data links and updated datasets as well :)


ljstella (approved):
LGTM

PR #2221, opened 2022-05-28
adding a note about how the score is calculated

josehelps:
Adds a simple note on how the RBA score is calculated 

image


ljstella (approved):
LGTM

PR #2241, opened 2022-06-08
Support custom indices in test files

pyth0n1c:
There are a number of detections whose
attack_data MUST be uploaded into specific
indices.  This is required due to macros that
are used to search for that data or for 
Datamodel Acceleration searches that require
data to be resident in specific indexes.
This PR adds the ability to specify a field
called "custom_index" in the test.yml files.
This field can be specified for each individual file
(just like source and sourcetype).  If this field is
not present, then a default index 
(at this time "main") will be used for replaying
data.


josehelps:
LGTM 😄

josehelps (approved):

PR #2279, opened 2022-07-08
Retry failed detections in cicd

pyth0n1c:
When a test fails, wait some time and retry the query a few times.
This may make truly failed tests take longer, but should
resolve the tests that are failing because we didn't wait
long enough for the data to be ingested.


pyth0n1c:
Small changes to CI/CD detection testing system. We merge these in now so that we can examine the results of the nightly scheduled run.

PR #2286, opened 2022-07-12
Check and dedup ids

pyth0n1c:
Added logic to contentctl.py to compare all IDs to ensure that a duplicate id is not used.  Also, resolved an issue where two detections had the same ID.  Right now, modifying the ID of an existing detection does not have any negative impacts.  

Now, when contentctl runs, if there are any collisions it will print out those collisions and return a nonzero exit code.  Please see the attached screenshot as an example of a collision.

The presence of an ID field is checked for ALL SecurityContentTypes, not just detections.  This means if any of those types have an ID added to them in the future, we will automatically check them.

image


patel-bhavin (approved):

patel-bhavin:
tested locally 👍

PR #2335, opened 2022-08-18
Grammar correction in documentation template

yaleman:
Just a quick tweak to the wording on the template.


patel-bhavin:
note: detection testing will fail since this PR coming from a fork. 
thanks @yaleman

PR #2346, opened 2022-08-29
Updated for XML format

TheLawsOfChaos:
Updated query to utilize XML format of Windows logs instead of Classic. Replaced the following fields:

- Message AS ScriptBlockText
- ComputerName AS Computer
- User AS UserID

Fixed in both the query and the stats line.

### Details

### Checklist

- [ ] Validated SPL logic.


patel-bhavin:
@TheLawsOfChaos : Hello, along with the search updates to use XML format we would also need the corresponding datasets in XML format for us to be able to test and merge these changes? Can you share the XML log file or the event which will trigger this search?
Thank you

patel-bhavin:
added this search update with new datasets to here : https://github.com/splunk/security_content/pull/2366

Issue #2342, opened 2022-08-27
missing $ on risk message


jsantoyo-r11:
the following detection (https://github.com/splunk/security_content/blob/develop/detections/endpoint/svchost_exe_lolbas_execution_process_spawn.yml) is missing the trailing $ for "$dest$" in the risk message tag.

swe:
Started a pull request to fix the issue

PR #2352, opened 2022-08-31
Update svchost_exe_lolbas_execution_process_spawn.yml

swe:
fixed line 53 by adding missing $

### Details

Fixed yml by adding missing character. This is a fix of issue #2342


patel-bhavin:
adding this update to a [new PR](https://github.com/splunk/security_content/pull/2366) since we currently don't support detection testing from forks. Thank you for fixing this

PR #2383, opened 2022-09-22
Support multiple tests per detection + HEC Replay

pyth0n1c:
This PR achieves a number of features and enhancements.  
Most important, it allows multiple tests to be listed (and separately
executed and logged) per detection Test file.  In order to support
this changes, there were a number of large updates to file
parsing, data replay, search execution, and summary output format:

- Detections/tests/results are now handled as objects instead of separate strings/dicts throughout detection testing
- Data replay is done via HEC Raw input with Indexer acknowledgement which is configured when the container is first started
- Summary files were updated to reflect pass/failures of multiple tests per detection instead of a single test


pyth0n1c:
This functionality will be added separately.

PR #2390, opened 2022-09-28
Add cis20 and nist tag checking

pyth0n1c:
Enhance validation for detection tags.
Improved validation for CIS20 tags
Added NIST tag validation


josehelps (approved):
LGTM

PR #2431, opened 2022-10-27
Updated search to use XML fields - malicious_powershell_executed_as_a_service

Rivosyke:
### Details

Changed the search block to reflect the XML fields. Like the PowerShell 4104s, the "user" field is now UserID and is only the user SID, which will need to be enriched (the TA doesn't do that).

Also updated the required_fields block.

### Checklist

- [ ] Validated SPL logic.

### DIFF
--- a/detections/endpoint/malicious_powershell_executed_as_a_service.yml
+++ b/detections/endpoint/malicious_powershell_executed_as_a_service.yml
@@ -8,12 +8,14 @@ datamodel:
 - Endpoint
 description: This detection is to identify the abuse the Windows SC.exe to execute
   malicious commands or payloads via PowerShell.
-search: ' `wineventlog_system` EventCode=7045 | eval l_Service_File_Name=lower(Service_File_Name)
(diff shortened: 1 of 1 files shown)


patel-bhavin:
@Rivosyke: thanks for updating these searches. In order for us to successfully merge the PR, we would like to know if you can provide attack_data for these detections so that we can ensure that our detections work as expected. Can you share the dataset you used to test these? 

Same for the other 3 PRs

josehelps:
bump .. @Rivosyke any way you can share a data dump/log file that we can use to test this detection with, for example the original was using https://media.githubusercontent.com/media/splunk/attack_data/master/datasets/attack_techniques/T1569.002/atomic_red_team/windows-system.log but obviously in multiline format.. a small (1-2) set of events that would trigger it would be sufficient. Otherwise, I can look into generating it ourselves.

patel-bhavin:
@Rivosyke : Closing this due to inactivity. Any update on the dataset? I will close the PR due to in activity. Please feel free to reopen this PR. Thank you

Issue #2422, opened 2022-10-21
SPL Bug - AWS Detect Users with KMS keys performing encryption S3


bowesmana:
This detection rule states a field requirement of requestParameters.bucketName

https://research.splunk.com/cloud/884a5f59-eec7-4f4a-948b-dbde18225fdc/

However, other than renaming that field, it's never used in the stats command.

I imagine this would make sense as a split by clause in addition to user, which would seem better than just lumping it into a general values() statement, however, if it's intended to be used, it's not, and if it's not intended, it should be removed from rename.

Either 
```
| stats count min(_time) as firstTime max(_time) as lastTime values(src_file) AS src_file values(dest_file) AS dest_file values(userAgent) AS userAgent values(region) AS region values(src) AS src by user bucketName
```
or 
```
| stats count min(_time) as firstTime max(_time) as lastTime values(bucketName) as bucketName values(src_file) AS src_file values(dest_file) AS dest_file values(userAgent) AS userAgent values(region) AS region values(src) AS src by user 
```

patel-bhavin:
thank you for reporting this, updated the SPL in this PR :  https://github.com/splunk/security_content/pull/2453

PR #2453, opened 2022-11-11
issue 2422 - fix spl

patel-bhavin:
updated spl according to https://github.com/splunk/security_content/issues/2422


josehelps (approved):
LGTM

PR #2455, opened 2022-11-11
Fix test naming

pyth0n1c:
This is a minor change that does not affect content at all.

A small number of tests contained an incorrect name for their related detections.
This fix has been made for the tests in question.


patel-bhavin:
LGTM!

patel-bhavin (approved):

PR #2460, opened 2022-11-15
add crowdstrike fdr to baseline

pyth0n1c:
adding the crowdstrike fdr app to the baseline
in order to support ingestion of crowdstrike fdr data:
https://splunkbase.splunk.com/app/5579


ljstella (dismissed):
LGTM

PR #2483, opened 2022-12-21
Change Process.process_name to Processes.process_name

sanjay900:
### Details
Found a small typo in a few searches, which was causing process_name to always be missing

### Checklist

- [x] Validate name matches `<platform>_<mitre att&ck technique>_<short description>` nomenclature
- [ ] [CI/CD](https://github.com/splunk/security_content/actions) jobs passed ✔️ 
- [x] Validated SPL logic.
- [x] Validated tags, description, and how to implement.
- [x] Verified references match analytic.

PR #2488, opened 2022-12-21
Process. - Processes.

patel-bhavin:
Updated SPLs from this PR-
https://github.com/splunk/security_content/pull/2483/files


tccontre (approved):
Thanks Bhavin for these good catch. LGTM

PR #2491, opened 2022-12-21
risk_index macro 

patel-bhavin:
This correlation rule can query from index=risk. Feedback via Slack :

https://splunk.slack.com/archives/C9RE1GUSZ/p1671648309339269


pyth0n1c (approved):
Looks good - macro renamed properly

PR #2503, opened 2023-01-06
Upload test artifact/results

pyth0n1c:
Uploads the summary results to the s3
reporting bucket after a test completes.
This step was in place before but never
ran because 1 or more tests fail.


dleung-splunk:
Might this PR get merged prior to the next release?  Just checking as I have some work that's dependent on the outcome of this, Thanks!  :)

pyth0n1c:
We have tagged it in preparation for the 3.56.0 release.  We expected to merge this tomorrow.

PR #2520, opened 2023-01-26
Attempted fix at "always" uploading

pyth0n1c:
Detection testing stage looks like it 
always passes, even if tests fail.
This is due to a change introduced to
ensure that for scheduled tests,
testing artifacts are uploaded to S3.

the if: always() action is not cancellable, so 
it cannot be "&&" with other conditions, making the
if logic challening.  We have attempted to update
it in line with the following suggestions:
https://github.com/actions/runner/issues/491#issuecomment-926924523


pyth0n1c:
Closed this out -  we will fix this a different way via a different branch

PR #2521, opened 2023-01-26
Revert "Upload test artifact/results"

patel-bhavin:
Reverts splunk/security_content#2503


pyth0n1c (approved):
Discussed with Bhavin.
We must make this changes because the
continue-on-error: true
line hides failures in our
detection testing workflow.
We will re-work how we decide
to upload test results to S3 in 
the coming weeks.

PR #2540, opened 2023-02-13
Remove "observables" from ss.conf annotations

patel-bhavin:
image

Removes observable from annotations.


pyth0n1c (dismissed):
Interactively tested this fix together last week.

pyth0n1c (approved):
Changes look good.  Tests updated so that automated testing via pytest works too!

PR #2574, opened 2023-03-17
Update permissions when committing

pyth0n1c:
Given the incoming change to default permissions for
the GITHUB_TOKEN, make a change that should address
this.  It will be challenging to test this change in this repo
without merging + tagging a release since that is the only
time the affected job is called.


ljstella (approved):
LGTM

PR #2612, opened 2023-04-11
Smoketest all detections

pyth0n1c:
A number of experimental and deprecated detections contained
syntactic, or other, errors that prevented them from running out of
the box.  While some of these are covered by the "How to Implement"
or "Description" fields (and may include downloading models or
installing specific apps), we will update any detections which cannot
run because of typos or other accidental oversights.  These changes
also create a job, which will run nightly, to smoketest all detections
so that we have insight into these failures in the future.

The smoketest executes the search SPL associated with each detection,
including **type: production**, **type: experimental**, and **type: deprecated**
and reports any errors that are returned. It does NOT replay data, even if it
is available, and does NOT check for the success of the detection.  That 
functionality is only performed against **type: production** detections.


patel-bhavin:
the workflow CI worked correctly the experimental and deprecated searches that fail the CI will be fixed seperately!
Thanks!

patel-bhavin (approved):

PR #2621, opened 2023-04-14
Aws Exfil Detections

patel-bhavin:
### Updated Analytic Story
- Suspicious AWS S3 Activities
- Data Exfiltration

### New Analytics

- AWS AMI Attribute Modification for Exfiltration
- AWS EC2 Snapshot Shared Externally
- AWS Exfiltration via Anomalous GetObject API Activity
- AWS Exfiltration via Batch Service
- AWS Exfiltration via DataSync Task
- AWS Exfiltration via EC2 Snapshot
- AWS Exfiltration via Bucket Replication
- AWS Disable Bucket Versioning
- AWS S3 Exfiltration Behavior Identified (Correlation)

image


detections/cloud/aws_disable_bucket_versioning.yml
mvelazc0, line 1: "Adversaries are seen to perform this activity during a ransomware incident in AWS." does not sound great. I would change this to something like: Adversaries have leverages this technique in the wild during a ransomware breach as describes in [link here]. Does not have to be exact, something like that

detections/cloud/aws_exfiltration_via_anomalous_getobject_api_activity.yml
mvelazc0, line 1: ...to detect anomalies with respect to users making high number of GetObject API calls to download objects from S3 by a user within 10 minutes of time window...

This is a bit confusing. It starts in plural "users making" but then goes to singular "by a user".

detections/cloud/aws_exfiltration_via_bucket_replication.yml
mvelazc0, line 1: I  think this should start like all our detections:
This search is for ...
The following analytic detects...

detections/cloud/aws_exfiltration_via_anomalous_getobject_api_activity.yml
patel-bhavin: updated to :
detect anomalies with respect to users making high number of GetObject API calls to download objects from S3 in a 10 minute time window.

detections/cloud/aws_disable_bucket_versioning.yml
patel-bhavin: updated:
Adversaries have leveraged this technique in the wild during a ransomware incident to disable versioning so the client cannot recover the data.

mvelazc0 (approved):

PR #2643, opened 2023-04-25
adding required fields

patel-bhavin:


pyth0n1c (approved):
Verified with dev that this should populate the field in the YML, but not interactively ask the user for input at the command line when creating the detection.

PR #2676, opened 2023-05-16
[TR-2961] Splunk Message Identifier Activity Analysis

ljstella:
### Details

First reviewable state for TR-2961

### Checklist

- [ ] Validated SPL logic.


kelby-shelton:
1. In my testing, my data source didn't have orig_recipient (stream:smtp). I recommend we switch to using | from datamodel since the split-by in TSTATS requires all fields be populated.
2. The output for value was showing up as a List for me. It looks like message_id is being assigned a list of values and we're not unpacking it. We need to pair the message ID with the recipient data. Its probably easier to ditch the playbook input and just eval the message ID in the splunk search.
3. Tag is missing D3-IAA
4. Description should include "Ref: [D3-IAA](https://d3fend.mitre.org/technique/d3f:IdentifierActivityAnalysis)" and explain that its using data in the Email data model.
5. Format block is missing the | --- | --- | --- | --- | underneath the header row. 

With those changes, I approve!

P4T12ICK (approved):
LGTM

PR #2686, opened 2023-05-23
KVstore update

patel-bhavin:
- Remove unused lookups
- Deleted some empty lookups and converted searches to use KV store
- Updated code to check for lookup validation


pyth0n1c:
The following commit introduces some changes that do the following:
make sure that a lookup.yml file contains either a filename of a collection, but NOT both.  If it does not, it will fail validation.
Makes sure that if a lookfile filename is specified, that the lookfile CSV actually exists in the lookups directory.
Finally, it ensures that every CSV file in the lookups directory is referenced (by one or more) lookup ymls.  

https://github.com/splunk/security_content/pull/2686/commits/cefd9349b161398522d599637cce77d98c5fca7b

pyth0n1c (approved):
Worked with the dev to resolve one minor issue.  Changes look good and pass validation and testing with no errors.

PR #2723, opened 2023-06-14
Adding Filter macros for  SSE

patel-bhavin:
This PR addes all the filter macros and other macros to the macros.json file via the API

image


pyth0n1c (approved):
Had a discussion with the author which resulted in minor cleanup.  Looks good to me, I approve.

PR #2735, opened 2023-06-22
Improved BA detection

P4T12ICK:
Fix a logical bug in the BA detection 
Windows Powershell Connect to Internet With Hidden Window


patel-bhavin (approved):

patel-bhavin:
The updated condition is correct now!

PR #2742, opened 2023-06-22
ASL search description 

patel-bhavin:
ASL Text reopen!


patel-bhavin:
i had to reopen this PR as i accidentally committed an error into develop!

Verified the text and it looks far better! LGTM!

PR #2743, opened 2023-06-22
new line check in description

patel-bhavin:
This PR checks to throw error in validate if an escape character is not present in the description and how to implement field when there is a new line!


pyth0n1c (approved):
Worked with PR author to make minor change in output string to be more descriptive. Tested and working!

PR #2746, opened 2023-06-27
Improve lookup file csv validation

pyth0n1c:
Currently, we rely on authors to make sure that the format of CSV files is correct.  However, there are a number of issues that can be introduced when creating or modifying existing CSV files that could introduce unintended behavior.  For example:
1.  Each row of a CSV file should have the proper number of columns (this should match the number of columns in the header)
2. Proper quoting should be enforced for all CSV files.  Failure to include proper quoting can result in the wrong number of columns in a row or improper formatting of a lookup table in-product


patel-bhavin (approved):

patel-bhavin:
LGTM, associated detections with lookups passed the CI!

PR #2766, opened 2023-07-11
G00t Duplicate for testing by Steven Dick

patel-bhavin:
@nterl0k : create a duplicate with updates to the yaml fields and for testing via CI


pyth0n1c (approved):
Had a discussion with PR author for minor fixes around field names for observable and making a search string more general.
All tests look good!
Approved.

PR #2834, opened 2023-08-31
updated BA spec

P4T12ICK:
Update BA spec


patel-bhavin:
merging this code into release_v4.12.0

PR #2852, opened 2023-09-18
Ba detections spl2

P4T12ICK:
BA detections in SPL2


patel-bhavin:
Detections are tested by @P4T12ICK. LGTM!

patel-bhavin (approved):

PR #2855, opened 2023-09-19
Sysmon searches updated to use CIM fields

patel-bhavin:
Updates all sysmon sourcetype searches to use CIM fields namely dest everywhere in SPL, message and observable"


ljstella (approved):
LGTM

Issue #2744, opened 2023-06-23
Improve performance of pretrained DGA model


dglauche:
The provided notebook has a logic flaw which results in time consuming unnecessary operations

I improved the code so that running it on thousand domains takes seconds not minutes: [link](https://github.com/splunk/security_content/compare/develop...dglauche:security_content:develop)

patel-bhavin:
@ksharad-splunk is taking a look at testing your proposed fix! Thank you @dglauche! nice hanging out at .conf

patel-bhavin:
The code has been tested and merged! Thank you @dglauche!

PR #2861, opened 2023-09-25
DGA Performance fix

nsreekanta:
This PR contains fix for DGA performance as suggested here https://github.com/splunk/security_content/issues/2744
Tested the fix and it greatly improves the performance


patel-bhavin (approved):

patel-bhavin:
The fix is tested by @nsreekanta and is similar to the proposed fix by @dglauche

PR #2881, opened 2023-10-07
Fixes for contentctl migration

pyth0n1c:
A number of issues were caught by the new contentctl codebase.

While fixes were made that "resolve" these errors, we should ensure that these were _actually_ errors and do not introduce regressions into the playbooks.


ljstella (approved):
Site still generates fine, looks the same to me. 

:shipit:

pyth0n1c:
> Site still generates fine, looks the same to me.
> 
> :shipit:

Based on this review, I have dropped the WIP / Work In Progress Tag

PR #2957, opened 2024-02-07
Removing trailing space from filename

codingWithJimmy:
- Trailing space from this filename causes issues with downstream deployments that utilize version control to manage updates in their environment

### Details

IHAC who uses version control to manage many different ES installations and the most recent update introduced an issue because of a trailing space in the filename. When trying to merge new content, git balks at the trailing space and breaks their workflows. Renaming the file to remove the trailing space to ensure this issue doesn't occur for others.

### Checklist

- [ x ] Validate name matches `<platform>_<mitre att&ck technique>_<short description>` nomenclature
- [ x ] [CI/CD](https://github.com/splunk/security_content/actions) jobs passed ✔️ 
- [ x ] Validated SPL logic.
- [ x ] Validated tags, description, and how to implement.
- [ x ] Verified references match analytic.


ljstella:
Thanks for reporting this @codingWithJimmy, we've opened an MR to get this fixed upstream of here, the next release should reflect it.

PR #2992, opened 2024-04-04
fix search and conditions in sigma version of w3wp_spawning_shell.yml

ajkingio:
### Details
/detections/endpoint/w3wp_spawning_shell.yml didn't translate correctly to Sigma. The condition logic was making so it would detect logs that only matched selection4 and selection5. It also wasn't looking for powershell.exe event though the descriptions it should be.
I've updated the search and logic to align with the description.

### Checklist

- [ ] Validated SPL logic.


patel-bhavin:
Hello @ajkingio : We currently DO NOT actively support/ maintain sigma search conversion for ESCU detection. If you see an error with the same associated in the `detections/` directory, please open up an issue / PR with requested changes. 

We will consider removing this dev/ directory completely since our detections work better when directly written into SPL which you can find in the detections/ . Thanks for the PR!

PR #2995, opened 2024-04-07
Nterl0k - [T1566++] - A bunch of O365 built-in / premium security content

nterl0k:
### Details

Mostly focused on bubbling up the various O365 security alerting for both built-in and premium features.

ZAP, DLP, Safe Links, Safe Attachments
Security & Compliance alerting
Report A Message function
Some insider threat behaviors for the 0365 platform.

Pending https://github.com/splunk/attack_data/pull/888 

### Checklist

- [ ] Validated SPL logic.


ljstella:
Okay, status update:
1) Handful of small tweaks made to each yaml for formatting and to pass validation. All set and passing `build` now. 
2) Currently, a single detection is failing to pass unit testing (O365 ZAP Activity Detection). The search actually works fine, but as part of that validation, we make sure that the fields specified in the `observable:` section (which gets translated into RBA) exists in the results. In the sample dataset, they don't all exist in all of the events. Two of the three events have a URL, and one of the three has a filename. Creating threat objects out of those is a great idea, I just need to confirm that it works properly in-product and that we don't have this limitation in the testing to guard against some other issue.

nterl0k:
Understandable... i'm probably causing all kinds of fun problems for you.

My "development" process is a bit different in that I design my submissions in a production Enterprise Security environment then once I've fire tested them a bit, I backport them to ESCU PRs

I'm sure that leaves some errors or odd qualities on them

Thanks for all the feedback!
________________________________

ljstella:
Follow up: currently blocking this PR on the completion of https://github.com/splunk/contentctl/pull/204

This, particularly the ZAP analytic and its dataset gave us a chance to deep dive on ES and the risk events created in it. The `src_user` field in that detection (the email address the email that was ZAP'd came from) was being listed as a risk object with type of Other, even though you had properly given it the Role of `Attacker`. This created different risk events than we would have necessarily expected. We're changing that so it works the way we'd expect. There'll likely be an overhaul of that part of our YAMLs entirely at some point in the future to use terms closer to what you'd see in ES and take away some of the mental overhead in that system, but for now, if you say something is an Attacker, it won't fall back to being a Victim without warning.

Secondly, and the reason this was failing testing- `contentctl` currently expects each Result (returned from running the detection) to have **all** of the fields in the Observable section. Essentially each Notable if that was configured for this detection would have to have all of them. That obviously doesn't work for this detection, and likely doesn't work for others, as an email could be ZAP'd because of a URL OR an attachment OR both, or likely some other reason too. We're looking to change `contentctl` so instead of each result having them, the field has to appear at least once in a set of all the fields. Once that's complete, this should start passing.

ljstella:
Status update: the required changes landed in [splunk/contentctl](https://github.com/splunk/contentctl)'s `main` branch yesterday. When the next release of contentctl happens, this will start passing CI.

ljstella (approved):
This is now passing with the release of contentctl 4.3.1 which allows for threat objects to be present in only some of the risk events.

Thanks again for this awesome contribution @nterl0k !

PR #2997, opened 2024-04-10
Wget/Curl Download and Bash Execution: Changing logic for search terms, to make searches comply with title and intention

DipsyTipsy:
### Details

The logic in the queries here has an OR between the matching on quiet/stdout and the match for pipe and "bash".
This causes the query to hit when wget/curl is either silenced or piped to bash.
From the title i would maybe assume that an AND between the quiet/stdout check and pipe+bash check would be the correct queries?

Searches modified:
- Wget Download and Bash Execution
- Curl Download and Bash Execution

### Checklist

- [ ] Validated SPL logic.


MHaggis:
Hey @DipsyTipsy , we are working to fix this up. I agree with your fixes. We're in the midst of ensuring we can generate new data that we can test against. We appreciate the feedback and apologize for the long delay in getting this resolved.

MHaggis:
Hey @DipsyTipsy , we got the updates updated. Apologies for the delay! This will go out next release along with a reference to you for your awesome help! Really appreciate the feedback and your patience!

patel-bhavin:
Hello @DipsyTipsy - sorry this took a bit for us to test, there are currently some TA limitations with the test data set but after testing this with @MHaggis , its evident that the detection needs to be fixed. We have added a note at the bottom of the yaml mentioning this limitation. Thank you for updating this detection !!

patel-bhavin (approved):

PR #3024, opened 2024-07-01
DLUX 1 - Adding Misc New Detections

dluxtron:
Adding a bunch of misc new detections across several security domains.

PR #3028, opened 2024-07-08
Unit-Testing using contentctl

patel-bhavin:
This PR enables Unit-testing in Github Actions CI using contentctl. 

Additionally added a script to pretty print the pass/failures of detections from summary.yml

image


.github/workflows/unit-testing.yml
```diff
@@ -0,0 +1,65 @@
...
+            pip install contentctl
+            git clone --depth=1 --single-branch --branch=master https://github.com/redcanaryco/atomic-red-team.git
+        
+        - name: Run ContentCTL test for changes against develop
+          run: |
+            git pull
+            source .venv/bin/activate
+            git checkout ${GITHUB_REF#refs/heads/}
+            # cat summary.yml
+            # contentctl test --post-test-behavior never_pause mode:changes --mode.target-branch develop
```
patel-bhavin: enable this before merging!

patel-bhavin:
Updated the output for better formatting in summary
image

.github/workflows/build.yml
```diff
@@ -23,29 +21,20 @@ jobs:
         run: |
```
pyth0n1c: I don't think this workflow needs jq (anymore) can it be removed entirely?

.github/workflows/unit-testing.yml
```diff
@@ -0,0 +1,63 @@
...
+            ref: develop
+
+        - uses: actions/setup-python@v5
+          with:
+            python-version: '3.11' #Available versions here - https://github.com/actions/python-versions/releases  easy to change/make a matrix/use pypy
+            architecture: 'x64' # optional x64 or x86. Defaults to x64 if not specified
+
+        - name: Install System Packages and contentctl 
+          run: |
+            sudo apt update -qq
```
pyth0n1c: Now that we are doing the processing with python i think we can remove jq?

pyth0n1c (dismissed):
I believe this looks good. After it is merged, we should do some stress-testing where we open a PR from an external (non splunk/security_content member) repo to ensure that the mode:changes functionality works from external forks as well.

.github/workflows/build.yml
patel-bhavin: removed!

.github/workflows/unit-testing.yml
patel-bhavin: removed

patel-bhavin:
This workflow was also tested on PR from Dean Luxton (Splunker) and it seemed to have worked correctly! You can check the run here : 

https://github.com/splunk/security_content/pull/3024

pyth0n1c (approved):
Reviewed workflow and results.  Looks great! Approved!

PR #3029, opened 2024-07-09
Remove BA yamls

patel-bhavin:
Removing BA yamls


ljstella:
Think you should set `build_ssa` to false in `contentctl.yml` before this merges- somewhat concerning that with it set to true but none of the files or folders present that contentctl doesn't yell.

patel-bhavin:
> Think you should set `build_ssa` to false in `contentctl.yml` before this merges- somewhat concerning that with it set to true but none of the files or folders present that contentctl doesn't yell.

Great point! I will push that change!

ljstella (approved):
This will break a bunch of links that are already likely already broken- otherwise this should be fine once we know what we're doing with it all.

PR #3030, opened 2024-07-09
Nterl0k - RMM Must Die - Update

nterl0k:
### Details

Added a global exception lookup (wildcards accepted) with filtering macro that interacts with the asset and identity model to provide a more uniform/easy way to suppress false positives/known usage of RMM tools in an environment. This content allows users to have one place to drop permanent (or temporary) allowances to the RMM content.

The lookup/macro also provides an optional date & days TTL for any items added.

This works in addition to the standard ESCU "filter" macros to provide a course and fine grain level of tuning for the content.

Exception List
image

Feilds added to existing rules by macro
image

### Checklist

- [ ] Validated SPL logic.


ljstella:
Testing is failing because the macro `remote_access_software_usage_exception_filter` calls a lookup `asset_lookup_by_str`- that's part of ES and not present in the testing containers. Checking internally to see how we want to handle testing this.

ljstella (dismissed):
This is now passing all testing and should be good to go!

Thanks again for your awesome contributions.

ljstella:
Update after discussion w/ @patel-bhavin:
- The `remote_access_software_exceptions` lookup was converted to a kvstore based lookup in order to ensure customer modifications are preserved in future updates of the app instead of being overwritten with the example data. 
- The `remote_access_software_exception_filter` macro was renamed, removing the `_filter` in order to more easily distinguish it from regular `_filter` macros that do not come pre-filled. 
- Because we switched to a kvstore based lookup from a CSV, we lost the example data that was pre-filling the exceptions lookup. Because of this, additional information was added to the `how_to_implement` sections of each detection that utilizes it. That doesn't feel quite as good but is . If you have a blog that references how to use it, adding it to the `references` section would be the easiest way to get the same if not more detail into the proximity of usage. 

That stuff aside, this is now passing all checks, and looks to be good to merge now!

ljstella (approved):

nterl0k:
Thinking I have a blog is giving me way too much internet credit. 😂
________________________________

patel-bhavin:
Thank you @ljstella and @nterl0k ! Snazzy updates in here :)

PR #3033, opened 2024-07-10
Fork testing

patel-bhavin:
@pyth0n1c :updated this code to enables unit testing to successfully execute on both normal PR and PRs created from forks.


pyth0n1c (approved):
Looks good, I have tested this from a separate repo as well.
We will still need to evaluate if testing work from an external repo where its source branch is named "develop" and a few other edge cases, but we should be mostly good!
We should also pursue not hardcoding the repository: security_content and ref: develop, if possible, to make cloning this repo and its workflows work as easily as possible for other users.

PR #3036, opened 2024-07-18
Updated precommit config

ljstella:
### Details

Updating precommit config base.

Discussion:
I'd like to add `detect-private-key` and `detect-aws-credentials` - I cannot imagine a scenario in which case we want to intentionally introduce either into this repo. I also think it would be rare that they would be accidentally introduced to this repo, as opposed to splunk/attack_range, although its not impossible. 

I'm also considering whether or not adding `forbid-submodules` would be helpful. While it would not conflict with any of our build processes, it would ensure we don't accidentally add them. 

### Checklist

- [ ] Validated SPL logic.


josehelps:
Any way to enforce the install of pre-commit hooks? Also or additionally should we add some docs on how and when to use them?

ljstella:
> Any way to enforce the install of pre-commit hooks? 

Not really, no. We'd need to use pre-recieve hooks instead, and that's... well, that's an entirely separate, bigger headache. 

> Also or additionally should we add some docs on how and when to use them?

We can definitely add some docs and/or a section to `README.md`. I think its probably about time we add a `CONTRIBUTING.md` and that would likely be the place to put it.

ljstella:
- Added 3 new hooks to pre-commit config

- Brought the Contribution Guidelines into the repo (.github/CONTRIBUTING.md) from the Project Wiki, updated filepaths and incorrect references to the repo name, and updated maintainers list. If we merge this, the wiki should have the page updated to reference this file.

ljstella:
Updated https://github.com/splunk/security_content/wiki/Developing-Content which is pointed to from `CONTRIBUTING.md`, it was still referencing CircleCI and schema generation for example. 

I _think_ we're good to go on this for now, but we really need to consider circling back to the Wiki and giving it a **thorough** cleaning.

patel-bhavin (approved):

PR #3042, opened 2024-07-23
remove unused lookup files

patel-bhavin:


pyth0n1c (approved):
Approved - a new pre-release version of contentctl determined that these files are no longer referenced and should be removed. This will soon be a blocking error in contentctl during validate.

PR #3047, opened 2024-07-23
Remove dev Sigma

patel-bhavin:
We do not actively support , update or use the the yamls. Leaving them in here often leads to confusion! This PR removes that.


ljstella:
@patel-bhavin can we tweak `.github/labeler.yml` to remove the `dev/` folder as part of this PR? This line in particular can just be removed with this change:
https://github.com/splunk/security_content/blob/625fc37fe292b146acbcdd887f35275085592499/.github/labeler.yml#L5

patel-bhavin:
done!

ljstella (approved):
LGTM

PR #3049, opened 2024-07-24
Better data source tagging

P4T12ICK:
Improved the data source handling:
- new simple data source schema (event source doesn't exist anymore)
- tagged a lot of detections which supported data sources
- data_source.csv and data source lookup
- data source field is part of annotations


patel-bhavin:
skipping results of unit-testing as no searches were changed. I will run a complete integration test on all these changes!

PR #3064, opened 2024-07-29
minor sysmon datasource update

patel-bhavin:
### Details

### Checklist

- [ ] Validated SPL logic.


ljstella (approved):
LGTM

:shipit:

PR #3070, opened 2024-07-30
Unit-test for forks

patel-bhavin:
We recently ran into an error with unit testing job for a `fork:develop` branch due to branch name collision when we switch into "develop"

This PR attempts to checkout the source branch of a fork into a new branch name : `new_branch_for_testing`

Sample test that works when the the PR is from fork:develop

https://github.com/splunk/security_content/pull/3071


pyth0n1c (approved):
This looks good to me - I have tested that this works locally and verified the testing in github as well

PR #3072, opened 2024-07-31
Improve TA names to match the naming in Splunk

P4T12ICK:
Changes:
- Add configuration file to Sysmon and Windows Event Code 4688
- Changes TA names in data sources to match the name in Splunk
- Updated TA version to match the latest (new check in contentctl)

PR #3082, opened 2024-08-05
Fix and Update Data Sources

pyth0n1c:
A few supported_ta fields were poorly
formatted in the new data_source objects.
Fix the formatting of those fields and
update those objects to use the latest
relevant TAs available on Splunkbase


pyth0n1c:
Didn't realize this work was already done: https://github.com/splunk/security_content/pull/3072
Closing out this PR

PR #3085, opened 2024-08-07
Updating risk_message

patel-bhavin:
updating the 3 detections that fail risk validation
update to contentctl.yml file with path fix


ljstella (approved):
LGTM

PR #3088, opened 2024-08-09
add missing version and id fields to story

pyth0n1c:
This story was missing the version and id fields.
This was allowed to happen due to a bug in contentctl which is tracked and fixed in the following branch: 
https://github.com/splunk/contentctl/pull/213


ljstella (approved):
:shipit:

PR #3091, opened 2024-08-13
Marked two detections as manual_test

pyth0n1c:
There is an issue tracking this internally.
 These detections currently fail validation of the Risk Message. We believe this is because the raw, exported data does not export the `event.EndpointIp` and `event.EndpointName` as required for the Risk Message to render properly.
However, we believe this is an artifact of the testing system and does not occur when data is ingested realtime in a production system.

### DIFF
--- a/detections/endpoint/crowdstrike_medium_severity_alert.yml
+++ b/detections/endpoint/crowdstrike_medium_severity_alert.yml
@@ -52,6 +52,10 @@ tags:
   - event.SeverityName
   risk_score: 49
   security_domain: endpoint
+  manual_test: This detection is marked manual test because the attack_data file and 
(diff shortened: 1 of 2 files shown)


ljstella:
is it `event.EndpointIp` and `event.EndpointName` or is it `src_host` and `src_ip` as indicated in the YAML?

pyth0n1c:
> is it `event.EndpointIp` and `event.EndpointName` or is it `src_host` and `src_ip` as indicated in the YAML?

src_host and src_ip are those fields, but renamed.
I have updated the YMLs and repushed to reflect the fixed manual_test field.

ljstella (approved):
:shipit:

PR #3092, opened 2024-08-15
Fix potential input output lookup issues

pyth0n1c:
Extended validation for lookups to include inputlookup and outputlookup.
This required a few content modifications and creation of new lookups.

Since we have a large number of detections that now use KVStore (for example, detections which use a baseline), convert those detections to write directly to the lookup_name (table) instead of lookup_name.csv


pyth0n1c:
Closing out this PR in favor of a new one. It has drifted too far from the baseline.

PR #3094, opened 2024-08-20
Remove Duplicate Obs

patel-bhavin:
Remove Duplicate Obs


patel-bhavin:
this change is included in https://github.com/splunk/security_content/pull/3061/

PR #3101, opened 2024-08-23
remove tab

tjgeorgen:
remove bad extra tab character


ljstella (approved):
LGTM, thanks for the catch

PR #3104, opened 2024-08-26
4.39.0 integration fixes

patel-bhavin:
UPdating SPL to user inner join to avoid unnecessary empty results


ljstella (approved):
LGTM

:shipit:

PR #3107, opened 2024-08-29
Improvements AWS ASL detection

P4T12ICK:
- bug fix detection ASL AWS Concurrent Sessions From Different Ips
- bump aws ta version


P4T12ICK:
I don't understand why datasource-ta-check still fails even though it passes locally.

patel-bhavin:
yep! confirmed that its passing locally and no idea why its showing failure on this PR.

patel-bhavin (approved):

PR #3114, opened 2024-09-03
Add data sources dependabot

patel-bhavin:
Remove the old data source CI file
Add a dependabot style CI job that updates data sources supported TA version on a nightly basis


P4T12ICK (dismissed):
looks good

PR #3117, opened 2024-09-04
README - Getting started

patel-bhavin:
Adding a getting started section
Quick start commands
Adding explanation for our yaml files


ljstella:
Couple of questions:

- Why are we putting them all in the readme.md as opposed to another file/folder and linking to it? 
- Do we think folks are going to continue to scroll that far to see them? 
- How will we ensure these stay up to date as the objects change?
- There are several other constraints on several of the fields in many files, that we've not indicated in these tables- are we planning to expand this more later, or just rely on validation failures to guide folks along?

patel-bhavin:
Updated - 

- added the spec files to the folder and linked them
- rearranged the contents of the yamls
- when a developer makes yaml change, it will be expected of them to update this spec file and document that change. If it needs to be automated - we can consider that in the future. 
- also added enums but we also rely on contentctl to guide a user through the errors

ljstella (approved):

patel-bhavin:
Known failure in app inspect! merging this PR

PR #3130, opened 2024-09-17
Application drilldowns

patel-bhavin:
Updated yaml structure after speaking with Eric:
All TTP, Anomaly and Correlation detections will have drilldown search associated with it 

```
drilldown_searches:
- name: View the detection results for $user$ and $dest$
  search: '%orginal_detection_search% | search  user = $user$ dest = $dest$'
```

`%orginal_detection_search%` -> this is written in such syntax so that we can dynamically update drilldown search when the seach is updated, and to keep the yaml clean

`%orginal_detection_search%`->  this field will be replaced by the search string during build time from contentctl  

Contentctl to make updates such that the following fields are optional in the yaml, if not present we use the following default, 
 ```
  ```


pyth0n1c:
FYI, I committed to this branch with the following fix.  
All the updated detections went from the typo
`%orginal_detection_search%` to
`%original_detection_search%`

PR #3133, opened 2024-09-18
Enrichments CI update

patel-bhavin:
Updates our CI jobs to run --enrichments from this PR 

https://github.com/splunk/contentctl/pull/291


pyth0n1c (approved):
I have released contentctl 4.3.5 and re-run the build job (which uses enrichments) and it looks like everything passes!
However, the appinspect workflow is currently failing. I believe this is because it is using the OLD, unupdated workflow file for inspect:
image

Unfortunately, this is just how the PR Target action works. As such, I am approving this PR even though that job appears to be failing.

PR #3139, opened 2024-09-24
Fix dates, fields with spaces, and update verisons for many detections

pyth0n1c:
In support of new versioning enforcement code and contentctl updates, clean up ymls and update version numbers and dates.
This is because the format/output for many detections in savedsearches.conf has changed slightly.


patel-bhavin:
This is a known failure due to some yamls not present in this repo. This app inspect bug will be fixed in the next releases

PR #3144, opened 2024-09-26
Update misnamed eventid

pyth0n1c:
A number of YMLs had issues around the data_source object(s) in the YMLs.
Either the wrong named were used for Sysmon EventID (a space was accidentally included), the key in the YML was `data_sources` instead of `data_source`, or other minor issues.
This PR should resolve those issues.

### DIFF
--- a/data_sources/windows_event_log_security_4688.yml
+++ b/data_sources/windows_event_log_security_4688.yml
@@ -1,7 +1,7 @@
 name: Windows Event Log Security 4688
 id: d195eb26-a81c-45ed-aeb3-25792e8a985a
-version: 1
-date: '2024-07-18'
+version: 2
+date: '2024-09-26'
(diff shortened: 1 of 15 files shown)


ljstella (approved):
LGTM-

Failing CI jobs are currently unavoidable.

ljstella (changes requested):
Does the change in `detections/application/windows_ad_privileged_group_modification.yml` require a version bump?

pyth0n1c:
> Does the change in `detections/application/windows_ad_privileged_group_modification.yml` require a version bump?

Yes, nice catch! In fact this IS caught by the version-checking, it's just obfuscated by the other error on missing non-public content.
https://github.com/splunk/security_content/actions/runs/11060840382/job/30732226304#step:5:246

image

ljstella (approved):
:shipit:

PR #3146, opened 2024-09-30
Add 2 drilldowns

patel-bhavin:
Adding two default drilldowns to all detections - TTP, Anomaly, TTP except experiemental/deprecated


ljstella:
Testing with the current `main` branch of contentctl produces the following validation errors: 

```
File: detections/endpoint/detect_critical_alerts_from_security_tools.yml
Error: 1 validation error for Detection
  Value error, This detection is required to have 2 drilldown_searches, but only has [0] [type=value_error, input_value={'name': 'Detect Critical...rom_security_tools.yml'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.9/v/value_error

File: detections/endpoint/windows_modify_registry_disable_rdp.yml
Error: 1 validation error for Detection
  Value error, This detection is required to have 2 drilldown_searches, but only has [0] [type=value_error, input_value={'name': 'Windows Modify ...gistry_disable_rdp.yml'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.9/v/value_error

File: detections/endpoint/windows_modify_registry_on_smart_card_group_policy.yml
Error: 1 validation error for Detection
  Value error, This detection is required to have 2 drilldown_searches, but only has [0] [type=value_error, input_value={'name': 'Windows Modify ..._card_group_policy.yml'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.9/v/value_error

VALIDATION FAILED
```

I'm not sure why we're requiring two drilldown searches- that _probably_ has to change, but in the meantime, these probably need to be updated. 

Let's sync when you're online about some of the testing that was done with Correlations specifically, since UI elements in ES are driven by their drilldown searches, which require specifically crafted searches.

pyth0n1c:
> Testing with the current `main` branch of contentctl produces the following validation errors:
> 
> ```
> File: detections/endpoint/detect_critical_alerts_from_security_tools.yml
> Error: 1 validation error for Detection
>   Value error, This detection is required to have 2 drilldown_searches, but only has [0] [type=value_error, input_value={'name': 'Detect Critical...rom_security_tools.yml'}, input_type=dict]
>     For further information visit https://errors.pydantic.dev/2.9/v/value_error
> 
> File: detections/endpoint/windows_modify_registry_disable_rdp.yml
> Error: 1 validation error for Detection
>   Value error, This detection is required to have 2 drilldown_searches, but only has [0] [type=value_error, input_value={'name': 'Windows Modify ...gistry_disable_rdp.yml'}, input_type=dict]
>     For further information visit https://errors.pydantic.dev/2.9/v/value_error
> 
> File: detections/endpoint/windows_modify_registry_on_smart_card_group_policy.yml
> Error: 1 validation error for Detection
>   Value error, This detection is required to have 2 drilldown_searches, but only has [0] [type=value_error, input_value={'name': 'Windows Modify ..._card_group_policy.yml'}, input_type=dict]
>     For further information visit https://errors.pydantic.dev/2.9/v/value_error
> 
> VALIDATION FAILED
> ```
> 
> I'm not sure why we're requiring two drilldown searches- that _probably_ has to change, but in the meantime, these probably need to be updated.
> 
> Let's sync when you're online about some of the testing that was done with Correlations specifically, since UI elements in ES are driven by their drilldown searches, which require specifically crafted searches.

The detections missing drilldowns have been updated - but this doesn't address your additional points. 
As you noted, let's have an offline chat about this and update the PR with our conclusions.

ljstella:
Current status of appinspect job: Its now targeting this branch appropriately so builds are actually happening, and should work across the board a bit better. The failures on that job are real. However, they may have been unintended- 

While hunting searches are excluded from the drilldown requirement, they are still getting a line added to their stanzas in savedsearches.conf that looks like this:
```
action.notable.param.drilldown_searches = []
```

This is because of the template changes made to add drilldown support. We can bump all those versions once, now, and keep this as a standard way we're going to output drilldowns, or we can go back to contentctl, change the template, try not to break anything, and not bump those versions.

Details about template:
This line is where we add them: https://github.com/splunk/contentctl/blob/cfda377c6887e28e02bb1798382ac0070b7983c2/contentctl/output/templates/savedsearches_detections.j2#L115

At first, I was a bit confused as to why it was there/what was going on. After loading it up in an IDE that identified pairs of brackets, I collapsed a few sections and found out the `{{if}}` block that it was inside was just the main one here at the top: https://github.com/splunk/contentctl/blob/cfda377c6887e28e02bb1798382ac0070b7983c2/contentctl/output/templates/savedsearches_detections.j2#L4

And that's how we got them on searches that don't require them.

ljstella (approved):
Should be fine to merge now. 

:shipit: I guess

PR #3162, opened 2024-10-15
fix misisng version for one analytic story

pyth0n1c:
Newly added analytic story was missing a field named `version`.
The presence of this field was not previously validated because, if it did not exist, it was added with a default of `1` at runtime.  That default value has been removed from the story type, so that if it is missing it is now a validation error.


ljstella (approved):
LGTM

PR #3168, opened 2024-10-18
add missing summariesonly macro

pyth0n1c:
We will still need to determine if we
want to add the missing COUNT after
it as well, but this will take a hands 
on test of the rule


patel-bhavin:
yes, lets add count and bump versions and date!

patel-bhavin (approved):

PR #3170, opened 2024-10-22
Add missing wildcard to -type parameter for detection 2452e632-9e0d-11eb-bacd-acde48001122

### LINKED ISSUE #3171: [BUG] Missing wildcard for -type parameter detection 2452e632-9e0d-11eb-bacd-acde48001122
The -type parameter doesn't have a leading wildcard, so it won't hit on e.g. "nslookup.exe -type=MX". 

### **Expected behavior**
It should also hit on nslookup.exe usage with the -type parameter. This will only work if the -type parameter has both a leading and trailing wildcard like the other parameters.

I've created a PR here:

https://github.com/splunk/security_content/pull/3170

Wouter-Jansen:
The detection won't hit on e.g. "nslookup.exe -type=MX" without the leading wildcard

### Details

This PR adds a leading wildcard to the -type parameter for nslookup.exe. Without that wildcard the -type parameter will never create a hit. All other parameters do have both a leading and trailing wildcard.

### Checklist

- [x] Validated SPL logic.

Closes #3171


patel-bhavin:
Awesome! Thank you for fixing this @Wouter-Jansen. As a request can you bump the version and the date of the detection yaml and please feel free to add your name to the detection as well since you help make it better!

ljstella (approved):
LGTM

:shipit: 

Thanks again for the contribution!

PR #3173, opened 2024-10-24
remove endhoursago=1 from drilldown searches

patel-bhavin:
Remove endhoursago=1 as it doesnt work on our testing. Rely on $info_max_time$ for the latest offset. Screenshot to show the _time span (highlighted)

Also update name for better formatting on Incident Review 

image


pyth0n1c (approved):
Despite the first apparent failure, the search `GPUpdate with no Command Line Arguments with Network` passes when retesting:
image

The second failed search `Potentially malicious code on commandline` passes after installing Python for Scientific Computing. I recommend adding this to our testing baseline. Note this this is a more time-consuming app to install. Here is evidence of the search passing:
image

PR #3175, opened 2024-10-24
Alerts detection update

patel-bhavin:
Updating detection to output mitre. 
 - eval the risk score based on severity and set it at search time
 - remove mitre from yml and set it at search time 

Data on Gamorr:
image


ljstella (approved):
¯\_(ツ)_/¯

:shipit:

PR #3179, opened 2024-10-31
Only for sysmon sources

patel-bhavin:
Updates sourcetype for only where['source'] == 'XmlWinEventLog:Microsoft-Windows-Sysmon/Operational'


ljstella (approved):
:shipit:

Issue #3134, opened 2024-09-19
[BUG] Azure MFA failure detections logic flaw


0xC0FFEEEE:
There are a number of Azure MFA failure based detections utilizing `properties.status.errorCode=500121` as a means of detecting one or more MFA auth failures. There's nothing wrong with this in of itself, however upon noticing a greater number of notables than anticipated with these detections enabled and performing some testing and research I have found the following from [here](https://learn.microsoft.com/en-us/entra/identity/monitoring-health/concept-sign-in-log-activity-details):

> Multifactor authentication: When a user signs in with MFA, several separate MFA events are actually taking place. For example, if a user enters the wrong validation code or doesn't respond in time, more MFA events are sent to reflect the latest status of the sign-in attempt. These sign-in events appear as one line item in the Microsoft Entra sign-in logs. **That same sign-in event in Azure Monitor, however, appears as multiple line items. These events all have the same correlationId.**

Note the last sentence.

Testing confirms that failed MFA does indeed result in more than one event for a single ocurrence of MFA auth. In addition to the above statement, it turns out that the `correlationId` can span multiple MFA auth attempts, however the `originalRequestId` field is unique for each attempt.

As an example of what this should look like, using `Azure AD Authentication Failed During MFA Challenge` as a basis for an updated search:

```
`azure_monitor_aad` category=SignInLogs properties.status.errorCode=500121 
| rename properties.* as * 
| stats count, min(_time) as firstTime, max(_time) as lastTime, values(user) as user, values(src_ip) as src_ip, values(mfaDetail.authMethod) as mfa_method, values(status.additionalDetails) as status.additionalDetails, values(reason) as reason, values(appDisplayName) as appDisplayName, values(user_agent) as user_agent by correlationId, originalRequestId
| `security_content_ctime(firstTime)` 
| `security_content_ctime(lastTime)` 
| `security_content_ctime(times)` 
| `azure_ad_authentication_failed_during_mfa_challenge_filter`
```

`correlationId` can be thought of as the login session, and `originalRequestId` the MFA attempt. `count` is effectively useless above, but is left in to hightlight the fact that multiple events are produced for each failed attempt.

It's also worth noting that there is inconsistent data returned by Azure AD. Multiple MFA failures .e.g. `MFA denied; user declined the authentication` can yield inconsistent/missing values in `status.additionalDetails` between attempts, and either the objectid or actual username (or both!) for the user field.

### **App Version:**
 - ESCU: 4.40.0

[...]

PR #3181, opened 2024-10-31
Azure AD Authentication Failed During MFA Challenge

0xC0FFEEEE:
### Details

Rewriting the  `Azure AD Authentication Failed During MFA Challenge` detection as documented in #3134.

Events are now grouped by `originalRequestId` which produces a more accurate representation of the sign-in event that can be observed within the Azure portal, particularly within the `Authentication Details` tab which the new `auth_detail` field seeks to replicate. In some circumstances there is actually now more detail available in this new field than is provided in the aforementioned tab within the Azure portal.

False positives have also been minimized by removing attempts that result in `MFA successfully completed` messages, which were found to be generated when a user opts to use a different MFA method than the default.

Further reductions in notable events can be achieved through filtering `MFA denied; duplicate authentication attempt` messages within the `auth_msg` field, as they could arguably be considered as false positives.

### Checklist

- [x] Validated SPL logic.
- [x] Verified references match analytic.


patel-bhavin:
thank for the contribution, the unit testing seems to fail due to some misformed Eval . I tested this latest search and if you can update the SPL to this , that would be great :

```
search : ->
`azure_monitor_aad` category=SignInLogs properties.status.errorCode=500121 
| rename properties.* as *, authenticationDetails{}.* as * 
| eval time=strptime(authenticationStepDateTime, "%Y-%m-%dT%H:%M:%S") 
| eval auth_detail=mvzip(strftime(time, "%Y-%m-%dT%H:%M:%S"),authenticationStepResultDetail," - "), auth_msg=mvappend(status.additionalDetails, authenticationStepResultDetail) 
| eval auth_method=mvmap(authenticationMethod, if(isnull(mvfind(mfaDetail.authMethod, authenticationMethod)), authenticationMethod, null())) 
| stats min(_time) as firstTime, max(_time) as lastTime, values(user) as user, values(src_ip) as src_ip, values(mfaDetail.authMethod) as mfa_method, values(auth_method) as auth_method, values(auth_detail) as auth_detail, values(auth_msg) as auth_msg, values(appDisplayName) as appDisplayName, values(user_agent) as user_agent by tenantId originalRequestId 
| search NOT auth_msg="MFA successfully completed" 
| sort 0 - firstTime 
| `azure_ad_authentication_failed_during_mfa_challenge_filter`
```

patel-bhavin:
Can you also update the text in the FP section to include this info? 

False positives have also been minimized by removing attempts that result in MFA successfully completed messages, which were found to be generated when a user opts to use a different MFA method than the default.

Further reductions in notable events can be achieved through filtering MFA denied; duplicate authentication attempt messages within the auth_msg field, as they could arguably be considered as false positives.

patel-bhavin:
@0xC0FFEEEE : thank you for updating this! I tested the evals locally and the PR looks good!

patel-bhavin (approved):

PR #3186, opened 2024-10-31
Powershell detection sourcetypes

patel-bhavin:


patel-bhavin:
the two failures is due to the attack data not being ingested correctly!

PR #3191, opened 2024-11-05
Updates to apps in contentctl to use latest

patel-bhavin:
- Removes latest from the link :
- Use the app that are used by attack range


josehelps (approved):
LGTM

PR #3197, opened 2024-11-06
Datasource dependabot feat. contentctl.yml

patel-bhavin:
This CI update helps maintain contentctl.yml file with up to date TAs, when something fails with the latest we can investigate and chose to revert to older version if need. 

The S3 bucket will be updated with the apps but the attack range depandabot.


P4T12ICK (dismissed):
Looks good to me

PR #3202, opened 2024-11-12
Strict yml fields

pyth0n1c:
Update content to begin enforcing STRICT YML fields in contentctl.
This means that we will throw errors when fields that are not defined in contentctl Pydantic Objects are included in YML files.

This reduces the chance that optional field names will be typo'd and cleans up a large number of YML files.

Note that the corresponding contentctl PR is here: https://github.com/splunk/contentctl/pull/325

To summarize the fields that have been removed:
Detection Tags:
- `risk_score` is no longer included in the YMLs.  This is calculated from `confidence` and `impact` fields
- `required_fields` have been removed
- `context` has been removed
- `group` has been removed
 
Baseline:
- `datamodel` has been removed 
- `required_fields` has been removed

Investigation
- `datamodel` has been removed

Many other unused fields have been removed (or, if they were misnamed, correctly to the proper name)


patel-bhavin:
These changes to the yamls look good. Also tested the associated contentctl PR  to check for potential validation errors created - looking good!

image

Also confirmed that the removed fields are not used in the product.

patel-bhavin:
OLD PR!

PR #3204, opened 2024-11-13
RBA Migration

ljstella:
### Details

Content migration accompanying changes to contentctl: https://github.com/splunk/contentctl/pull/263

### Checklist

- [ ] Validated SPL logic.


ljstella:
Status update:

I have (temporarily) changed the GitHub Actions config in my PR (https://github.com/splunk/contentctl/pull/263) to run the ESCU smoketest against this branch. 

Here are the latest results: https://github.com/splunk/contentctl/actions/runs/11860779327/job/33057262480?pr=263

As of this point, there are now under 100 validation errors against translated content. 

The vast majority of these will need to be handled individually. Up until this point, commits have generally been done by subdirectory of `detections/` with all of the changes for those in each one to assist in review. If you're reviewing this change at a later point, this is a great time to go refill your coffee ☕  and grab a snack 🍿 . I will attempt to group similar fixes and/or subdirectories together from this point forward, but they may be a bit more spread out.

ljstella:
Update: as of now, **all** translated content is building successfully: https://github.com/splunk/contentctl/actions/runs/11863423975?pr=263

Still have to do some cleanup around messages and some risk objects, but they're all technically passing and producing a package "compatible" with the previous releases. I say compatible, as opposed to identical, because as this work was done, some changes have been made where things were threat objects but probably should have been risk objects, vice versa, and some where they just didn't make sense as either and were removed. Some risk messages were updated along the way, although there will be more of that upcoming.

cmcginley-splunk:
This isn't a fully formed thought, but considering the recent discussions we've had around the notables and risk analysis events in ES 8.0, do we think it makes sense to lump notables under the terminology `risk_objects`? Does it make sense to explicitly distinguish risk analysis objects from notable objects in the YAML? I'm not convinced it does, just wanna raise a convo around it

ljstella:
@cmcginley-splunk 
> do we think it makes sense to lump notables under the terminology risk_objects?

Could you expand on this a bit? I don't think we have anything that does this today, but generally speaking (at least, pre-ES8) notables were **extremely** simple, especially in comparison to risk events, in form of schema, required bits and bobs, and most importantly, a simple 1:1 relationship between search results and journaled events. 

Some of the drivers behind these changes like aligning the terminology in the YAML with that in the product, and a more simplistic representation of the objects in config with fewer levels of abstraction seem at odds with the idea of putting notable-related things into an `rba` object.

Do I think that at a future point, we should consider a similar effort with a `notable` object that configures all of the `action.notable.param` fields based on the detection's deployment? Perhaps, although we can sort of already do that with the in-line deployments.

patel-bhavin:
@ljstella : Shall we update the github workflows build, inpect and test to install the alpha version of contentctl?

ljstella:
> @ljstella : Shall we update the github workflows build, inpect and test to install the alpha version of contentctl?

Once there is an alpha release of contentctl 5.0.0, these are the spots ~it needs to change~ (actually, maybe doesn't need to change with that): 
 
Here: 
https://github.com/splunk/security_content/blob/863eb07dc93edfa87c594f5c0fbbedf670e032b2/.github/workflows/build.yml#L22

Here:
https://github.com/splunk/security_content/blob/863eb07dc93edfa87c594f5c0fbbedf670e032b2/.github/workflows/unit-testing.yml#L26

Here: https://github.com/splunk/security_content/blob/863eb07dc93edfa87c594f5c0fbbedf670e032b2/.github/workflows/appinspect.yml#L21

Here: 
https://github.com/splunk/security_content/blob/863eb07dc93edfa87c594f5c0fbbedf670e032b2/.github/workflows/datasource-dependabot.yml#L27

patel-bhavin:
App inspect changes works locally ! LGTM

patel-bhavin (approved):

PR #3207, opened 2024-11-14
update version and date

patel-bhavin:
update version and date


pyth0n1c (approved):
All bumps look good to me and so does appinspect with metadata validation.
Approved!

PR #3213, opened 2024-11-26
Enhancing Coverage - First Batch

nasbench:
This PR aims to update some detection with additional coverage as well as enforces the use of some already defined macros where they were missing. Below is an explanation of the additions with some details. Feel free to reach out if you need more details.

- PowerShell 7 names it's process "pwsh.exe". So it just make sense to add it to detections covering powershell.exe already (for the most part)
- Added `7zr.exe` to the analytic `01d29b48-ff6f-11eb-b81e-acde48001123` as it's also a valid option to use along the other versions of 7zip
- Added additional ways of how the administrator is written/represented in other language packs in the analytic `b89919ed-fe5f-492c-b139-151bb162040e`
- Added the expanded version of the alias "sl" which is "set-log" to 236e7c8e-c9d9-11eb-a824-acde48001122
- Added coverage for the ability to call comsvc with ordinal #24 to dump creds to 8943b567-f14d-4ee8-a0bb-2121d4ce3184
- Added additional option for the analytic c148a894-dd93-11eb-bf2a-acde48001122 covering more feature that you can disable with defender
- Made the search for the backup keyword broader in cd5aed7e-5cea-11eb-ae93-0242ac130002 as `backup` is also a legitimate (hidden) flag that allows the deletion of backups.
- Excluded some default cases for sdbinst usage to avoid triggering on any execution (details below)
- Enforced usage of the `process_nltest` macro for analytics calling nltest
- Enforced usage of the `process_net` macro for analytics calling net or net1
- Enforced usage of the `process_wmic` macro for analytics calling wmic
- Fixed typo in "Environment" 

#### Sdbinst Exclusion

The following flags and strings were excluded in the sdbinst rule

- `-?` because we don't care about people using it with the help flag
- Excluded both `C:\\Windows\\System32\\sdbinst.exe` and `"C:\\Windows\\System32\\sdbinst.exe"` with quotes to avoid simply the execution of the binary with command line options, as it is an unworthy event in this specific case. It can be useful if sdbinst is used for process injection as a child of an uncommon process (but that would fit in another rule)
- Excluded when the CLI ends with "-mm" as that's a default behavior on windows 11 and above coming from a builtin task - Read [this](https://github.com/nasbench/Misc-Research/blob/main/Other/Undocumented-Flags-Sdbinst.md) for more details
{9915DED8-8441-4D39-9FD3-A02AF5D018AB}


pyth0n1c (approved):
Manually reviewed changes and confirmed successful testing.
Looks good!

PR #3222, opened 2024-12-02
Update linux_auditd_file_permission_modification_via_chmod.yml

MHaggis:
Update based on feedback from Ivar Nygård
Thank you!

TR-3940


nasbench (approved):
LGTM for this fix, but imo we should discuss this rule internally to perhaps enhance it, as some of the permissions like 755, 754 and 700 aren't suspicious or malicious. We could perhaps add a version with suspicious paths to reduce the FP rate that "will" occur

cc @tccontre

pyth0n1c:
> LGTM for this fix, but imo we should discuss this rule internally to perhaps enhance it, as some of the permissions like 755, 754 and 700 aren't suspicious or malicious. We could perhaps add a version with suspicious paths to reduce the FP rate that "will" occur
> 
> cc @tccontre

@MHaggis and I chatted about some of these this morning.  We are thinking the best approach might be a very verbose, new detection with a number of `case` statements and different regexes, explaining what each of those different cases are.
We already have a number of chmod detections (although they are against a number of different log sources), but a larger "catch-all" would be really useful.

PR #3233, opened 2024-12-09
fix #2996

### LINKED ISSUE #2996: [BUG] Missing Wildcards in Splunk Rule for Detecting Known Services Killed by Ransomware
Product: Splunk Security Content

Description:

The Splunk rule named "known_services_killed_by_ransomware" is intended to detect when certain critical services have entered the stopped state, potentially as a result of ransomware activity. However, there is an inconsistency in the use of wildcards within the Message IN list, which may lead to the rule not triggering on relevant log messages.

Steps to Reproduce:
Navigate to the Splunk search interface.
Enter the provided Splunk rule named "known_services_killed_by_ransomware".
Execute the rule against the wineventlog_system log data containing EventCode 7036.

Expected Result:

The rule should match any Message that contains the names of the specified services followed by the phrase "service entered the stopped state", regardless of any additional text or variations in case.

Actual Result:

Due to missing wildcards for some service names and the lack of a comma between "QBCFMonitorService" and "YooBackup", the rule may not match all relevant Message field values, resulting in potential false negatives.

Bug Impact:

The effectiveness of the rule in detecting service interruptions related to ransomware activities is compromised. Critical alerts may be missed, reducing the ability of the security team to respond to ransomware threats in a timely manner.

Suggested Fix:

Review and update the rule to ensure consistent use of wildcards with all service names in the Message IN list. Add missing wildcards where necessary to capture variations in the Message field values. Ensure that all service names are correctly separated by commas.

nasbench:
### Details

This PR aims to fix the issues raised by #2996 by adding the appropriate missing comma and removes the overlapping "BackupExecVSSProvider" and "BackupExecAgentAccelerator" that are already covered by "BackupExec*"

The list of services was gathered from https://news.sophos.com/en-us/2020/04/24/lockbit-ransomware-borrows-tricks-to-keep-up-with-revil-and-maze/ hence most of these do not need wildcards (as the op assumed).

### DIFF
--- a/detections/endpoint/known_services_killed_by_ransomware.yml
+++ b/detections/endpoint/known_services_killed_by_ransomware.yml
@@ -1,7 +1,7 @@
 name: Known Services Killed by Ransomware
 id: 3070f8e0-c528-11eb-b2a0-acde48001122
-version: '6'
-date: '2024-11-28'
+version: 7
+date: '2024-12-09'
 author: Teoderick Contreras, Splunk
 status: production
 type: TTP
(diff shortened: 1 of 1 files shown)


nasbench:
Looping @tccontre just because its your detection.

tccontre (approved):
LGTM

PR #3237, opened 2024-12-12
Improved ASL AWS detections

P4T12ICK:
This PR updates the searches and datasets for the latest version of Amazoin Security Lake.


patel-bhavin:
@P4T12ICK : We should add a data source files for ASL, that way we can also better clarify these detections are using Cloudtrail logs from ASL. Pushed some minor updates to the yamls!

P4T12ICK:
Good feedback. I added the data source and referenced it in the detections.

patel-bhavin (approved):

PR #3240, opened 2024-12-13
crypto_campaign

tccontre:
### New Detection

- [x] windows_powershell_logoff_user_via_quser.yml
- [x] windows_obfuscated_files_or_information_via_rar_sfx.yml
- [x] windows_remote_management_execute_shell.yml
- [x]  windows_file_and_directory_permissions_enable_inheritance.yml
- [x] windows_file_and_directory_permissions_remove_inheritance.yml
- [x] windows_account_access_removal_via_powershell_initiate_logoff.yml
- [x] windows_bidirectional_communication_via_telegram_bot.yml
- [x] windows_file_and_directory_enable_readonly_permissions.yml

### updated
- [x] icacls_grant_command.yml
- [x] executables_or_script_creation_in_suspicious_path.yml
- [x] suspicious_process_file_path.yml

### tagged

- [x] add_or_set_windows_defender_exclusion.yml
- [x] any_powershell_downloadfile.yml
- [x] attempt_to_stop_security_service.yml
- [x] chcp_command_execution.yml
- [x] cmd_carry_out_string_command_parameter.yml
- [x] detect_password_spray_attack_behavior_on_user.yml
- [x] detect_rare_executables.yml
- [x] download_files_using_telegram.yml
- [x] excessive_service_stop_attempt.yml
- [x] excessive_usage_of_cacls_app.yml
- [x] excessive_usage_of_sc_service_utility.yml
- [x] excessive_usage_of_taskkill.yml
- [x] hiding_files_and_directories_with_attrib_exe.yml
- [x] high_process_termination_frequency.yml
- [x] icacls_deny_command.yml
- [x] malicious_powershell_process___encoded_command.yml
- [x] modify_acl_permission_to_files_or_folder.yml
- [x] permission_modification_using_takeown_app.yml
- [x] sc_exe_manipulating_windows_services.yml
- [x] suspicious_scheduled_task_from_public_directory.yml
- [x] windows_account_access_removal_via_quser_logoff.yml
- [x] windows_autoit3_execution.yml
- [x] windows_boot_or_logon_autostart_execution_in_startup_folder.yml
- [x] windows_disable_or_modify_tools_via_taskkill.yml
- [x] windows_obfuscated_files_or_information_via_rar_sfx.yml
- [x] windows_remote_management_execute_shell.yml
- [x] windows_screen_capture_in_temp_folder.yml
- [x] windows_service_created_with_suspicious_service_path.yml
- [x] windows_service_creation_using_registry_entry.yml
- [x] windows_service_deletion_in_registry.yml
- [x] windows_service_stop_by_deletion.yml
- [x] windows_service_stop_via_net__and_sc_application.yml
- [x] windows_system_file_on_disk.yml
- [x] windows_system_user_discovery_via_quser.yml
- [x] xmrig_driver_loaded.yml

### story
        new file:   stories/crypto_stealer.yml

### Checklist

- [ ] Validated SPL logic.


detections/endpoint/windows_account_access_removal_via_quser_logoff.yml
```diff
@@ -0,0 +1,66 @@
...
+type: TTP
+status: production
+description: The following analytic detects the process of logging off a user through the use of the quser and logoff commands. By monitoring for these commands, the analytic identifies actions where a user session is forcibly terminated, which could be part of an administrative task or a potentially unauthorized access attempt. This detection helps identify potential misuse or malicious activity where a user’s access is revoked without proper authorization, providing insight into potential security incidents involving account management or session manipulation.
+kind: endpoint
+search: '`powershell` EventCode=4104 ScriptBlockText = "*quser*" ScriptBlockText = "*logoff*" 
+  | stats count min(_time) as firstTime max(_time) as lastTime by EventCode  ScriptBlockText UserID Computer 
+  | rename Computer as dest, UserID as user 
+  | `security_content_ctime(firstTime)` 
+  | `security_content_ctime(lastTime)` 
+  | `windows_account_access_removal_via_quser_logoff_filter`'
```
nasbench: This logic can cause some to a lot of FPs. You have 2 option here.

The first is you could reflect the logic using quser and then iterating and logging them off by focusing on everything that is on a single line like this. `*quser*logoff*`. This ofc can be bypassed and covers only some variations.

The other is to focus on the process creation version, Since calling any of the aforementioned commands will call their respective binaries auser.exe and logoff.exe

Depending on what you choose to do, it will either be an anomaly or hunting and should have a reduced score.

In its current form it is unusable.

detections/endpoint/windows_obfuscated_files_or_information_via_rar_sfx.yml
```diff
@@ -0,0 +1,68 @@
...
+version: 1
+date: '2024-12-12'
+author: Teoderick Contreras, Splunk
+data_sources:
+- Sysmon Event ID 11
+type: TTP
+status: production
+description: The following analytic detects the creation of RAR Self-Extracting (SFX) files by monitoring the generation of file related to rar sfx installation. This method leverages a heuristic to identify RAR SFX archives based on specific markers that indicate a combination of executable code and compressed RAR data. By tracking such activity, the analytic helps pinpoint potentially unauthorized or suspicious file creation events, which are often associated with malware packaging or data exfiltration. Legitimate usage may include custom installers or compressed file delivery.
+kind: endpoint
+search: '`sysmon` EventCode=11 Image="*.exe" TargetFilename IN ("*__tmp_rar_sfx_access_check*")
```
nasbench: You don't need the `Image="*.exe"` as the filename in itself is indicative.
 
```suggestion
search: '`sysmon` EventCode=11 TargetFilename IN ("*__tmp_rar_sfx_access_check*") 
```

detections/endpoint/windows_obfuscated_files_or_information_via_rar_sfx.yml
```diff
@@ -0,0 +1,68 @@
+name: Windows Obfuscated Files or Information via RAR SFX
+id: 4ab6862b-ce88-4223-96c0-f6da2cffb898
+version: 1
+date: '2024-12-12'
+author: Teoderick Contreras, Splunk
+data_sources:
+- Sysmon Event ID 11
+type: TTP
```
nasbench: From VT data we can see that the creation of such file is pretty common
image


Hence putting it as Hunting is wise.
```suggestion
type: Hunting
```

detections/endpoint/windows_obfuscated_files_or_information_via_rar_sfx.yml
```diff
@@ -0,0 +1,68 @@
+name: Windows Obfuscated Files or Information via RAR SFX
```
nasbench, line 1: The logic is looking for the creation of that file. So the title needs to reflect that.

I didn't do full testing but from what i could gather online is that this file will get created every time an SFX is executed hence the title should be something like `Windows SFX Temporary File Creation` or something along these lines.

```suggestion
name: Windows Obfuscated Files or Information via RAR SFX
```

detections/endpoint/windows_obfuscated_files_or_information_via_rar_sfx.yml
```diff
@@ -0,0 +1,68 @@
...
+- name: View risk events for the last 7 days for - "$dest$"
+  search: '| from datamodel Risk.All_Risk | search normalized_risk_object IN ("$dest$") starthoursago=168  | stats count min(_time) as firstTime max(_time) as lastTime values(search_name) as "Search Name" values(risk_message) as "Risk Message" values(analyticstories) as "Analytic Stories" values(annotations._all) as "Annotations" values(annotations.mitre_attack.mitre_tactic) as "ATT&CK Tactics" by normalized_risk_object | `security_content_ctime(firstTime)` | `security_content_ctime(lastTime)`'
+  earliest_offset: $info_min_time$
+  latest_offset: $info_max_time$
+tags:
+  analytic_story:
+  - Crypto Stealer
+  asset_type: Endpoint
+  confidence: 50
+  impact: 50
```
nasbench: Based on the comments above the score should also be reduced significantly

nasbench (changes requested):

detections/endpoint/windows_obfuscated_files_or_information_via_rar_sfx.yml
tccontre: thank you man for the review. great point, currently the format of naming convention we have is < OS ><Mitre Technique/Tactic><Name>. also we have limited file name length I think :)

detections/endpoint/windows_obfuscated_files_or_information_via_rar_sfx.yml
tccontre: thanks man! ;)

detections/endpoint/windows_obfuscated_files_or_information_via_rar_sfx.yml
tccontre: noted! regarding the rar sfx the reason why I tag it TTP because it is not commonly used in production as a software installer. usually it was seen as unwanted third part utility tool installed by user

detections/endpoint/windows_account_access_removal_via_quser_logoff.yml
tccontre: thanks bro. the detection was fixed and renamed to powershell. I also checked that parent child process thing, unfortunately since the parent process is powershell that execute the script, it is powershell-> logoff.exe not so unique but worth to add will create another one and we can tag as anomaly ;)

detections/endpoint/windows_obfuscated_files_or_information_via_rar_sfx.yml
tccontre: as we've discuss offline converted it to anomaly ;)

detections/endpoint/windows_obfuscated_files_or_information_via_rar_sfx.yml
nasbench: Its noted. We'll discuss a bit more internally.

detections/endpoint/windows_account_access_removal_via_powershell_initiate_logoff.yml
```diff
@@ -0,0 +1,70 @@
...
+data_sources:
+- Sysmon EventID 1
+- Windows Event Log Security 4688
+type: Anomaly
+status: production
+description: The following analytic detects the process of logging off a user through the use of the quser and logoff commands. By monitoring for these commands, the analytic identifies actions where a user session is forcibly terminated, which could be part of an administrative task or a potentially unauthorized access attempt. This detection helps identify potential misuse or malicious activity where a user’s access is revoked without proper authorization, providing insight into potential security incidents involving account management or session manipulation.
+kind: endpoint
+search: '| tstats `security_content_summariesonly` min(_time) as firstTime max(_time) as lastTime from datamodel=Endpoint.Processes 
+  where Processes.parent_process_name = powershell.exe AND Processes.process_name = logoff.exe
+  by Processes.parent_process_name Processes.parent_process Processes.process_name Processes.process Processes.process_guid Processes.dest Processes.user
```
nasbench: I think its best to make this into a generic rule. Simply calling logoff.exe
Focusing on powershell is nice but removing it would not change the threat scope.

In my opinion having the rule just logoff.exe execution regardless of who called it is an interesting anomaly (in some env)

detections/endpoint/windows_bidirectional_communication_via_telegram_bot.yml
```diff
@@ -0,0 +1,67 @@
+name: Windows Bidirectional Communication via Telegram Bot
```
nasbench: I know of the convention but the title is confusing and does not reflect the logic. The rule is looking for initiated DNS query request. So it should be something that reflect that.

detections/endpoint/windows_bidirectional_communication_via_telegram_bot.yml
```diff
@@ -0,0 +1,67 @@
...
+id: 86f66f44-94d9-412d-a71d-5d8ed0fef72e
+version: 1
+date: '2024-12-12'
+author: Teoderick Contreras, Splunk
+data_sources:
+- Sysmon EventID 22
+type: Anomaly
+status: production
+description: The following analytic detects the use of DNS API calls associated with Telegram, which may indicate access via a Telegram bot commonly used by malware for command and control (C2) communications. By monitoring DNS queries related to Telegram's infrastructure, the detection identifies potential attempts to establish covert communication channels between a compromised system and external malicious actors. This behavior is often observed in cyberattacks where Telegram bots are used to receive commands or exfiltrate data, making it a key indicator of suspicious or malicious activity within a network.
```
nasbench:
```suggestion
description: The following analytic detects the execution of a DNS query by a process to the associated Telegram API domain, which could indicate access via a Telegram bot commonly used by malware for command and control (C2) communications. By monitoring DNS queries related to Telegram's infrastructure, the detection identifies potential attempts to establish covert communication channels between a compromised system and external malicious actors. This behavior is often observed in cyberattacks where Telegram bots are used to receive commands or exfiltrate data, making it a key indicator of suspicious or malicious activity within a network.
```

detections/endpoint/windows_bidirectional_communication_via_telegram_bot.yml
```diff
@@ -0,0 +1,67 @@
...
+kind: endpoint
+search: '`sysmon` EventCode=22  query = "api.telegram.org" 
+  | stats count min(_time) as firstTime max(_time) as lastTime  by query answer QueryResults QueryStatus process_name process_guid Computer 
+  | rename Computer as dest 
+  | `security_content_ctime(firstTime)` 
+  | `security_content_ctime(lastTime)` 
+  | `windows_bidirectional_communication_via_telegram_bot_filter`'
```
nasbench: To be confirmed but I believe The logic needs to exclude the telegram binary on windows because it communicates with the API itself. 
I don't know if there is a direct link from the web but excluding browsers is will also avoid FPs

detections/endpoint/windows_file_and_directory_enable_readonly_permissions.yml
```diff
@@ -0,0 +1,74 @@
...
+author: Teoderick Contreras, Splunk
+data_sources:
+- Sysmon EventID 1
+- Windows Event Log Security 4688
+type: TTP
+status: production
+description: The following analytic detects instances where file or folder permissions are modified to grant read-only access. Such changes are characterized by the presence of read-related permissions (e.g., R, REA, RA, RD) and the absence of write (W) or execute (E) permissions. Monitoring these events is crucial for tracking access control changes that could be intentional for restricting access or indicative of malicious behavior. Alerts generated by this detection help ensure that legitimate security measures are enforced while unauthorized changes are promptly investigated.
+kind: endpoint
+search: '| tstats `security_content_summariesonly` min(_time) as firstTime max(_time) as lastTime from datamodel=Endpoint.Processes 
+  where Processes.process_name IN( "icacls.exe", "cacls.exe", "xcacls.exe") AND Processes.process IN ("*/grant*", "*/G*") AND Processes.process IN ("*SYSTEM*", "*admin*", "*S-1-1-0*", "*EVERYONE*")
```
nasbench, line 13: Is there a specific reason we only focus on these groups for read only? I think any file being set to read only is interesting.

nasbench (changes requested):
A couple more changes before we can merge this.

detections/endpoint/windows_account_access_removal_via_powershell_initiate_logoff.yml
tccontre: thank you! fixed!

detections/endpoint/windows_bidirectional_communication_via_telegram_bot.yml
tccontre: fixed!

patel-bhavin (approved):

nasbench (approved):

PR #3262, opened 2025-01-02
Automated Splunk TA Update 133

patel-bhavin:

### DIFF
--- a/contentctl.yml
+++ b/contentctl.yml
@@ -77,9 +77,9 @@ apps:
 - uid: 5579
   title: Splunk Add-on for CrowdStrike FDR
   appid: Splunk_TA_CrowdStrike_FDR
-  version: 2.0.2
+  version: 2.0.3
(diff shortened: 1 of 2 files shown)


ljstella:
Most of these failures are for a `SearchParser` error- I can replicate _some_ of them locally, but not necessarily consistently. My hunch at the moment is that some of these searches are actually starting before all of the apps are installed. Running locally and logging into the splunk instance, some of these failures start popping **very** early in the full test run, before all the apps are visible in the UI.

Also, not sure we're propagating the errors correctly, as some of these are being listed as failing with the `SearchParser` error for `` `security_content_summariesonly` `` when the search itself doesn't use that macro.

ljstella (approved):
Pinning container to 9.3 as it avoids what appears to be a race condition between app installation and testing starting. 

Single known failure in results that I think I have a solution for soon.

PR #3267, opened 2025-01-06
Nterl0k - T1033 Query.exe usage on remote devices.

nterl0k:
### Details

More specific detection of Query.exe being used against remote devices.

Pending https://github.com/splunk/attack_data/pull/931

### Checklist

- [ ] Validated SPL logic.


detections/endpoint/windows_system_remote_discovery_with_query.yml
```diff
@@ -0,0 +1,57 @@
+name: Windows System Remote Discovery With Query
+id: 94859172-a521-474f-97ac-4cf4b09634a3
+version: 1
+date: '2025-01-06'
+author: Steven Dick
+status: production
+type: TTP
```
nasbench: Having query simply execute is not TTP worthy imo.

```suggestion
type: Anomaly

```

detections/endpoint/windows_system_remote_discovery_with_query.yml
```diff
@@ -0,0 +1,57 @@
...
+id: 94859172-a521-474f-97ac-4cf4b09634a3
+version: 1
+date: '2025-01-06'
+author: Steven Dick
+status: production
+type: TTP
+description: The following analytic detects the execution of `query.exe` with command-line arguments aimed at discovering data on remote devices. It leverages data from Endpoint Detection and Response (EDR) agents, focusing on process names and command-line executions. This activity is significant as adversaries may use `query.exe` to gain situational awareness and perform Active Directory discovery on compromised endpoints. If confirmed malicious, this behavior could allow attackers to identify various details about a system, aiding in further lateral movement and privilege escalation within the network.
+data_source: 
+- Sysmon Event ID 1
+- Windows Security Event ID 4688
```
nasbench, line 11:
```suggestion
- Windows Security Event ID 4688
- CrowdStrike ProcessRollup2
```

detections/endpoint/windows_system_remote_discovery_with_query.yml
```diff
@@ -0,0 +1,57 @@
...
+how_to_implement: The detection is based on data that originates from Endpoint Detection and Response (EDR) agents. These agents are designed to provide security-related telemetry from the endpoints where the agent is installed. To implement this search, you must ingest logs that contain the process GUID, process name, and parent process. Additionally, you must ingest complete command-line executions. These logs must be processed using the appropriate Splunk Technology Add-ons that are specific to the EDR product. The logs must also be mapped to the `Processes` node of the `Endpoint` data model. Use the Splunk Common Information Model (CIM) to normalize the field names and speed up the data modeling process.
+known_false_positives: Administrators or power users may use this command for troubleshooting.
+references:
+- https://attack.mitre.org/techniques/T1033/
+tags:
+  analytic_story: 
+  - Active Directory Discovery
+  asset_type: Endpoint
+  confidence: 75
+  impact: 60
```
nasbench: A single event for query is medium severity at best

```suggestion
  impact: 49

```

detections/endpoint/windows_system_remote_discovery_with_query.yml
```diff
@@ -0,0 +1,57 @@
...
+| `windows_system_remote_discovery_with_query_filter`'
+how_to_implement: The detection is based on data that originates from Endpoint Detection and Response (EDR) agents. These agents are designed to provide security-related telemetry from the endpoints where the agent is installed. To implement this search, you must ingest logs that contain the process GUID, process name, and parent process. Additionally, you must ingest complete command-line executions. These logs must be processed using the appropriate Splunk Technology Add-ons that are specific to the EDR product. The logs must also be mapped to the `Processes` node of the `Endpoint` data model. Use the Splunk Common Information Model (CIM) to normalize the field names and speed up the data modeling process.
+known_false_positives: Administrators or power users may use this command for troubleshooting.
+references:
+- https://attack.mitre.org/techniques/T1033/
+tags:
+  analytic_story: 
+  - Active Directory Discovery
+  asset_type: Endpoint
+  confidence: 75
```
nasbench: it should be medium https://github.com/splunk/security_content/wiki/6.1-%E2%80%90-How-are-risk-score-calculated-for-RBA

```suggestion
  confidence: 49

```

detections/endpoint/windows_system_remote_discovery_with_query.yml
```diff
@@ -0,0 +1,57 @@
...
+  - Splunk Enterprise Security
+  - Splunk Cloud
+  required_fields: 
+  - _time
+  - Processes.dest
+  - Processes.user 
+  - Processes.process
+  - Processes.process_name 
+  - Processes.parent_process_name
+  risk_score: 45
```
nasbench:
```suggestion
  risk_score: 24

```

detections/endpoint/windows_system_remote_discovery_with_query.yml
```diff
@@ -0,0 +1,57 @@
...
+known_false_positives: Administrators or power users may use this command for troubleshooting.
+references:
+- https://attack.mitre.org/techniques/T1033/
+tags:
+  analytic_story: 
+  - Active Directory Discovery
+  asset_type: Endpoint
+  confidence: 75
+  impact: 60
+  message: The user $user$ ran the Query command on a remote system from $dest$
```
nasbench:
```suggestion
  message: The user $user$ ran the Query command to query information from a remote system from $dest$

```

nasbench (changes requested):

detections/endpoint/windows_system_remote_discovery_with_query.yml
```diff
@@ -0,0 +1,57 @@
+name: Windows System Remote Discovery With Query
```
nasbench, line 1: Just noticed that we also have ad03bfcf-8a91-4bc2-a500-112993deba87 focusing on user discovery. In an edge case where a user would use 

`query user /sever:` they would both match creating an inflated score.

We perhaps need to exclude the "/server" flag from the other rule to focus only on local queries

nterl0k: Agreed, I noticed the duplicate firing in our environment and choose to alter risk scoring on the user discovery correlation accordingly. 

I can agree this is a bit of an overlap in coverage with the existing detection, but feel the lateral movement aspect adds more risk.

detections/endpoint/windows_system_remote_discovery_with_query.yml
nterl0k: Could vary depending on environment, but I have no issues setting it to RBA only.

nterl0k:
This would pass if I didn't put the testing data in the wrong folder 

https://github.com/splunk/attack_data/pull/936

detections/endpoint/windows_system_remote_discovery_with_query.yml
nasbench: Altered the detection of both rules to not overlap.

nasbench (approved):

PR #3280, opened 2025-01-15
Nterl0k - T1114.003 O365 Transport Rule Changed

nterl0k:
### Details

Pretty simple detection for changes to O365 Exchange transport rules.

Pending https://github.com/splunk/attack_data/pull/944

### Checklist

- [ ] Validated SPL logic.


detections/cloud/o365_email_transport_rule_changed.yml
```diff
@@ -0,0 +1,67 @@
...
+- name: View risk events for the last 7 days for - "$user$"
+  search: '| from datamodel Risk.All_Risk | search normalized_risk_object IN ("$user$") starthoursago=168  | stats count min(_time) as firstTime max(_time) as lastTime values(search_name) as "Search Name" values(risk_message) as "Risk Message" values(analyticstories) as "Analytic Stories" values(annotations._all) as "Annotations" values(annotations.mitre_attack.mitre_tactic) as "ATT&CK Tactics" by normalized_risk_object | `security_content_ctime(firstTime)` | `security_content_ctime(lastTime)`'
+  earliest_offset: $info_min_time$
+  latest_offset: $info_max_time$
+- name: Investigate changes by $user$ 
+  search: '`o365_management_activity` Workload=Exchange AND Operation IN ("Set-*","Disable-*","New-*","Remove-*") AND Operation="*Transport*" UserId=$user$'
+  earliest_offset: $info_min_time$
+  latest_offset: $info_max_time$
+rba:
+  message: The user [$user$] altered the exchange transport rule id $object_id$ [$object_name$]
```
patel-bhavin: should be - The user [$user$] altered the exchange transport rule id [$object_name$]

patel-bhavin (approved):

detections/cloud/o365_email_transport_rule_changed.yml
patel-bhavin: updated!

patel-bhavin:
@nterl0k :minor update in here and the PR looks solid! Thank you for the analytic!

PR #3283, opened 2025-01-17
Nterl0k - T1200 - Are you down with USB ?

nterl0k:
### Details

A set of detections aimed at tracking attachment of removable media and process execution from removable media (USB). Uses registry key changes to monitor for USB presence.

Pending https://github.com/splunk/attack_data/pull/952

### Checklist

- [ ] Validated SPL logic.


detections/endpoint/windows_detect_process_executed_from_removable_media.yml
```diff
@@ -0,0 +1,81 @@
+name: Windows Detect Process Executed From Removable Media
```
patel-bhavin: Can we update the name to `Windows Process Executed From Removable Media` . I think we should remove detect from all the detections in this! We are trying to be a bit better about naming!

detections/endpoint/windows_detect_process_executed_from_removable_media.yml
```diff
@@ -0,0 +1,81 @@
+name: Windows Detect Process Executed From Removable Media
+id: b483804a-4cc0-49a4-9f00-ac29ba844d08
+version: 1
+date: '2025-01-17'
+author: Steven Dick
+status: production
+type: Anomaly
+description: This analytic is used to identify when a process is executed with a removable media device as it's current drive or run path. Adversaries and Insider Threats may use removable media devices for several malicious activities, including initial access, execution, and exfiltration.
```
patel-bhavin: This is a neat analytic. Can you add more details to the description in terms of how and why object_handle is extracted and joined with Registry changes?

detections/endpoint/windows_detect_process_executed_from_removable_media.yml
nterl0k: Done.

detections/endpoint/windows_detect_process_executed_from_removable_media.yml
nterl0k: Done.

patel-bhavin:
@nterl0k : Thank you for updating the text! The PR looks good!

patel-bhavin (approved):

PR #3286, opened 2025-01-21
Version bumps

patel-bhavin:
Some detections needs version bumps from the latest app inspect runs


pyth0n1c (approved):
Just bumps versions on affected content. 
Approved.

PR #3291, opened 2025-01-23
analytics_enhancement

tccontre:
### renamed
        
- [x] known_services_killed_by_ransomware.yml -> detections/endpoint/windows_security_and_backup_services_stop.yml
- [x]  suspicious_process_file_path.yml -> detections/endpoint/windows_suspicious_process_file_path.yml
        

### updated
- [x] linux_auditd_find_credentials_from_password_managers.yml
- [x]  permission_modification_using_takeown_app.yml
- [x] suspicious_driver_loaded_path.yml
- [x] windows_driver_load_non_standard_path.yml

### new macro
- [x] macros/normalized_service_binary_field.yml

### new detections
- [x] windows_process_execution_in_temp_dir.yml

Screenshot 2025-01-22 at 15 12 59

### TTP to Anomaly
- [x] linux_auditd_file_permissions_modification_via_chattr.yml
- [x] linux_auditd_hidden_files_and_directories_creation.yml
- [x] linux_auditd_service_started.yml
- [x] linux_auditd_stop_services.yml

### Checklist

- [ ] Validated SPL logic.

### DIFF
--- a/detections/endpoint/windows_suspicious_process_file_path.yml
+++ b/detections/endpoint/windows_suspicious_process_file_path.yml
@@ -1,0 +1,122 @@
+name: Windows Suspicious Process File Path
+id: 9be25988-ad82-11eb-a14f-acde48001122
+version: 7
+date: '2025-01-27'
+author: Teoderick Contreras, Splunk
+status: production
+type: TTP
+description: The following analytic identifies processes running from file paths not
+  typically associated with legitimate software. It leverages data from Endpoint Detection
+  and Response (EDR) agents, focusing on specific process paths within the Endpoint
+  data model. This activity is significant because adversaries often use unconventional
+  file paths to execute malicious code without requiring administrative privileges.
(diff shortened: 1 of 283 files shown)


detections/endpoint/windows_suspicious_process_file_path.yml
```diff
@@ -1,7 +1,7 @@
-name: Suspicious Process File Path
+name: Windows Suspicious Process File Path
```
ljstella, line 1: This rename needs a deprecation/new detection, otherwise you're deleting the old detection and supplying a new one all at once.

ljstella (commented):
One rename needs another look, its currently deleting the existing detection and creating a new one

detections/endpoint/windows_suspicious_process_file_path.yml
tccontre: thanks @ljstella .I already fixed this issue.

ljstella (approved):
LGTM

PR #3287, opened 2025-01-22
Add descriptions and Mitre components to data sources

delgado-jacob:
### Details

We're integrating this data into the SnapAttack platform and I added some additional elements to facilitate that process. Would love to have this saved back to the repo, if possible.
- Expanded the description verbiage to try to capture what events the data source encompasses
- Added a mapping to Mitre Att&ck Data Source component(s) that the log would fall under
- Added a field to capture the separator_value to complement the separator
- Added a few new data source entries for the other types of bro/zeek logs (renamed the existing `Bro` source to `Bro http` since it was limited to http sourcetype)
- I did not increment the version or dates since I don't know what criteria is being used for determining a version change, but I can do that if appropriate.

PR #3301, opened 2025-01-30
output normalization for AWS cloudtrail logs

P4T12ICK:
output normalization for AWS cloudtrail logs


ljstella:
Started https://github.com/splunk/contentctl/pull/361 to track the code changes for this. Changing the CI to use that branch will clear the errors for extra fields. 

There is/will be conflict at some point with the PR from the Snapattack folks here: https://github.com/splunk/security_content/pull/3287 and the associated contentctl changes to accommodate that. 

I _think_ this looks good so far but there's a lot of questions about how we're going to use these that might make us want to change this up a bit.

ljstella (approved):
Think this is fine for now- no mechanism for enforcement/validation yet (code is merged, but no release with it yet so this CI hasn't ran with it)

PR #3303, opened 2025-01-30
Update Lookups & Windows EventLog Macros

nasbench:
This PR updates the windows event log macros for better consistency and coverage, as well as some lookups with additional data. Below are details of the updates:

### asr_rules

Added 2 additional rules that are in preview to the list, based on https://learn.microsoft.com/en-us/defender-endpoint/attack-surface-reduction-rules-reference#asr-rule-to-guid-matrix

### builtin_groups_lookup

Added additional groups based on our research on SDDL - https://github.com/MHaggis/SDDLMaker/tree/main/MindMap

### dynamic_dns_providers_default

Re-ordered the list alphabetically, and added a couple more entries

### security_services_lookup

Added multiple entries to other AV and EDR vendors based on https://github.com/Pennyw0rth/NetExec/blob/c904f3aa813f36f56b61d1021d32f6b527b20d0a/nxc/modules/enum_av.py and https://github.com/netwrix/pingcastle/blob/3e377c62f143ddf4db8c871a10bdf7c8b4605406/Scanners/AntivirusScanner.cs
and make the lookup search a case-insensitive search.

## Macro Updates

- All windows event log macros have been updated to use both `WinEventLog` and `XmlWinEventLog` .
- Added `PowerShellCore/Operational` to the PowerShell macro.
- Fixed the WMI macro from using `sourcetype` to using the `source` (similar to all other macros).
- Added `channel` logic to the built-in event logs (Application, System, Security).


patel-bhavin:
Nice clean up! LGTM

patel-bhavin (approved):

patel-bhavin:
Transient attack data failure for one detection, I believe we can ignore this error! 
image

patel-bhavin (approved):

PR #3309, opened 2025-02-05
Inspect 5.0.1

patel-bhavin:
Version bumps


pyth0n1c (approved):
Good idea to make contentctl version a CI variable so that it can be resused.
This now passes inspect validation when --suppress-missing-content-warnings is used.


Because the appinspect CICD job runs as `pull_request_target`, it is presently showing failure but this is just an aritifact of the CI system. As such, it can be ignored at this time until this PR is merged.

PR #3310, opened 2025-02-05
Ctl for forks!

patel-bhavin:
Updating to the way we install contentctl in actions for PRs and forks!


pyth0n1c (approved):
this works well for me locally after changing github style variable substitution to bash style substitution.

Approved!

PR #3323, opened 2025-02-10
Clean up MITRE Tagging

pyth0n1c:
When a type is defined, such as T1003, DO NOT allow a subtype such as T1003.001 to be defined. Remove the generic type T1003 and keep the subtype T1003.001. However, it is acceptable for a subtype to be defined or for a type to be defined separately. It is also okay for multiple subtypes to be defined.


patel-bhavin (approved):

patel-bhavin:
The changes on a high level looks good, this clean up helps in being more accurate about the exact  technique the detection is looking for

PR #3329, opened 2025-02-12
upating appinspect command

patel-bhavin:
Updating inspect command


pyth0n1c (approved):
Looks good, I should have caught this earlier as I knew the command line changed.

PR #3344, opened 2025-02-19
change match to two underscores

pyth0n1c:
use two underscores in filename to identify dashboard in line with
https://github.com/splunk/contentctl/pull/373


patel-bhavin:
this fixes the app inspect bug and the UI renders fine for github builds :
image

patel-bhavin:
inspect also checks out good!

patel-bhavin (approved):

PR #3347, opened 2025-02-19
Remove Incompatible 4688 Data Sources

nasbench:
This PR removes the 4688 data source from analytics that leverage `Processes.parent_process` since EventID 4688 does not have a `ParentCommandLine` field.


patel-bhavin:
looks good!

patel-bhavin (approved):

PR #3348, opened 2025-02-19
Update Exchange Macro

nasbench:
This PR updates the Exchange macro to the actual sourcetype defined by the TA https://splunkbase.splunk.com/app/3225 specifically the `TA-Windows-Exchange-IIS`

@inspired was the one to notice this during XS24, so kudos to him.

Since there is only one analytic currently using. I took the liberty to rename the macro to a more accurate name. I do not think this will affect anyone in a major way, but please keep me honest. If needed i could revert the rename.

image
image


patel-bhavin:
yeah, its better shipped as with windows_exchange and I checked this is not used by any detection other than the experimental change you made! thnak you

patel-bhavin (approved):

patel-bhavin:
also, wow! that's a lot of sourcetype configs in there :jenkins:

PR #3350, opened 2025-02-20
Minor fixes for integration tests

patel-bhavin:
Minor fixes for integration tests -  5.1.0

### DIFF
--- a/detections/endpoint/windows_service_created_with_suspicious_service_name.yml
+++ b/detections/endpoint/windows_service_created_with_suspicious_service_name.yml
@@ -12,7 +12,7 @@ search: |-
   `wineventlog_system` EventCode=7045 
   | stats values(ImagePath) as process, count, min(_time) as firstTime, max(_time) as lastTime values(EventCode) as signature by Computer, ServiceName, StartType, ServiceType, UserID
   | eval process_name = mvindex(split(process,"\\"),-1)
-  | rename Computer as dest, ServiceName as object_name, ServiceType as object_type, UserID as user
+  | rename Computer as dest, ServiceName as object_name, ServiceType as object_type, UserID as user_id
   | lookup windows_suspicious_services service_name as object_name
   | where isnotnull(tool_name)
   | `security_content_ctime(firstTime)` 
@@ -44,7 +44,7 @@ rba:
   - field: dest
     type: system
     score: 75
-  - field: user
+  - field: user_id
     type: user
     score: 75
   threat_objects: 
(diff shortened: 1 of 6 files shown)


detections/endpoint/windows_service_created_with_suspicious_service_name.yml
```diff
@@ -5,14 +5,14 @@ date: '2025-02-07'
...
 data_source: 
 - Windows Event Log System 7045
 search: |-
   `wineventlog_system` EventCode=7045 
-  | stats values(user) as user, values(ImagePath) as process, count, min(_time) as firstTime, max(_time) as lastTime values(EventCode) as signature by Computer, ServiceName, StartType, ServiceType, UserID
+  | stats values(ImagePath) as process, count, min(_time) as firstTime, max(_time) as lastTime values(EventCode) as signature by Computer, ServiceName, StartType, ServiceType, UserID
   | eval process_name = mvindex(split(process,"\\"),-1)
-  | rename Computer as dest, ServiceName as object_name, ServiceType as object_type
+  | rename Computer as dest, ServiceName as object_name, ServiceType as object_type, UserID as user_id
```
ljstella, line 15: if we're renaming UserID (as I think we should), shouldn't we opt for the more-CIM-compatible rename to `user`?

patel-bhavin: will update!

patel-bhavin: i made the update to user but I think user_id is also a CIM compatible field - https://docs.splunk.com/Documentation/CIM/6.0.2/User/CIMfields

ljstella: user_id is probably fine then *if* this is an actual UserID and not username.

The issue is that the userID to user name mapping is likely not happening in A&I, and so these will end up as rather weird dangling risk objects that aren't associated with the rest of the risk associated with the actual user that likely ties to a user name.

pyth0n1c: The value of this field the parsing spits out is from
`<Security UserID='S-1-5-21-2690122726-1172718210-436210976-500'/`
which results in the value being
`S-1-5-21-2690122726-1172718210-436210976-500`

ljstella: Yeah, that _would_ be a user_id in `Authentication` - not sure if that's what we actually want for a risk object long term though

patel-bhavin: only dest is attributed to risk :)

ljstella (approved):
:shipit:

PR #3355, opened 2025-02-21
Updating search output! 

patel-bhavin:
-  Improving the output of Windows InstallUtil Remote Network Connection
- Also added a data source object for Splunk CIM : addition of this object will help with keeping the contentctl.yml up to date. 
- We can also consider adding this object to all data model based searches!

### DIFF
--- a/detections/endpoint/windows_installutil_remote_network_connection.yml
+++ b/detections/endpoint/windows_installutil_remote_network_connection.yml
@@ -1,6 +1,6 @@
 name: Windows InstallUtil Remote Network Connection
 id: 4fbf9270-43da-11ec-9486-acde48001122
-version: 11
+version: 11 
 date: '2025-02-22'
 author: Michael Haag, Splunk
 status: production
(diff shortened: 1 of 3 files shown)


detections/endpoint/windows_installutil_remote_network_connection.yml
```diff
@@ -16,15 +16,24 @@ description: The following analytic detects the Windows InstallUtil.exe binary m
...
-  process process_id dest_port C2 | `windows_installutil_remote_network_connection_filter`'
+search: |-
+  | tstats `security_content_summariesonly` count FROM datamodel=Endpoint.Processes
+    where `process_installutil` by _time span=1h Processes.process_id Processes.process_name
+    Processes.dest Processes.user Processes.process_path Processes.process Processes.parent_process_name
+    Processes.original_file_name 
+  | `drop_dm_object_name(Processes)` 
+  | `security_content_ctime(firstTime)` 
+  | `security_content_ctime(lastTime)` 
+  | join process_id
```
ljstella: I realize this didn't change, but are we sure joining only on `process_id` is enough? Specifically because `process_id` is the simple number PID (as opposed to the `process_guid` that we see on the Endpoint model, but is missing from the Network_Traffic model). We might want to consider joining on `process_id` and `dest`, which will limit collisions to PID reuse on the same device, as opposed to every device logging the data.

patel-bhavin:
Updated the join! 

image

ljstella (approved):
LGTM

PR #3356, opened 2025-02-21
fix more typo'd data source names

pyth0n1c:
Fix typos in the names of data_sources.

Unsure if it will make it into the upcoming release, so I have NOT updated the versions (yet).
If this goes into 5.2, we will need to bump the versions.


patel-bhavin:
Closing this in favour of : https://github.com/splunk/security_content/pull/3357

PR #3357, opened 2025-02-21
Add data source objects

patel-bhavin:
Adding / fixing data source objects so that we can change the warning to errors on contentctl validate


P4T12ICK (approved):
LGTM

PR #3360, opened 2025-02-24
Automated Splunk TA Update 186

patel-bhavin:

### DIFF
--- a/contentctl.yml
+++ b/contentctl.yml
@@ -143,9 +143,9 @@ apps:
 - uid: 1876
   title: Splunk Add-on for AWS
   appid: Splunk_TA_aws
-  version: 7.9.0
+  version: 7.9.1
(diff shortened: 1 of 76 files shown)


ljstella:
Bunch of apps being updated now. 

Failed during the test job because of an error downloading 

```
https://attack-range-appbinaries.s3.us-west-2.amazonaws.com/splunk-add-on-for-amazon-web-services-aws_791.tgz
```

This will need to be manually adjusted to the correct URL (Another instance where the attack-range automation uses a different name than this automation does)

patel-bhavin:
Uploaded to S3 : seems like they have fixed that app file name on splunkbase . This is the latest file name name that gets downloaded from Splunkbase : splunk-add-on-for-amazon-web-services-aws_791.tgz

PR #3366, opened 2025-02-26
Remove 151 Deprecated detections, stories, baselines, investigations - Github

patel-bhavin:
- Moving to the top level removed/ directory. 
- adding deprecation_mapping.yml (includes 2 splunk detections
- adding semi - manually generated csv from the mapping yml (to be fully automated in next release)


pyth0n1c (approved):
I am approving. I discussed with @patel-bhavin and we came to a few points of agreement:
1. If ONLY a status is changed for a file which does not use ES8 versioning, like baselines, a version bump does not need to happen.
2. When something is moved from a deprecated/ folder into removed/ NOTHING should be updated. This includes the date and the version.
3. There were a number of baselines we failed to tag for removal in 5.2.0.  However, the detections which use them WERE tagged appropriately for removal. Since they are not uselful without the detections they support, they have been moved directly from production->removed without spending 2 releases in the deprecated state.

PR #3369, opened 2025-02-27
Updating dcsync detection due to false positive

dluxtron:
A customer found this detection can be triggered using the following command. 
dcdiag.exe /Test:Replications /s:DCNAMEHERE

This is still a replication event initiated by a user, but its not a dcsync attack.  
Updating the search to remove the logon status & adding details to the known false positives.


patel-bhavin:
discussed the changes with the author over Slack! Thank you @dluxtron

patel-bhavin (approved):

PR #3311, opened 2025-02-06
Automated Splunk TA Update 168 - Failing Auth detection

patel-bhavin:
This PR contains updates to Splunk TAs made by GitHub Actions workflow.

This detection is currently failing due to a bug in the Linux TA

### Issue :
The TA version 10.0.0 is currently failing on one of the detections in ESCU that uses Authentication datamodel and relies on data with `sourcetype = linux_secure` - https://github.com/splunk/security_content/pull/3311:

Here is the [detection](https://github.com/splunk/security_content/blob/3eaad9a5cded09e10d78b16e7b69f381cab19275/detections/endpoint/potential_password_in_username.yml#L4) 

- We believe that the addition of `AND NOT sourcetype=linux_secure` to the eventtypes searches is causing some data to not be mapped correctly to the Authentication Datamodel. 

Image
 
What is the reason to add this - [AND NOT sourcetype=linux_secure](https://github.com/splunk/splunk-add-on-for-unix-and-linux/pull/597/files#diff-73c9f338dcc47742b895fff359a09df3d2bf0830aa64e972b6b9085e7a91cc79R246) ? 

### What add-on version are you using?

10.0.0

PR #3390, opened 2025-03-13
Automated Splunk TA Update 203

patel-bhavin:


ljstella:
Screenshot 2025-03-13 at 8 55 10 AM

I have no idea wtf is happening to these version numbers, or why that's now causing the automation to try and downgrade the Appdynamics TA.

Since that's the only change other than the one we're working on via #3311, going to close this and see if it comes back.

PR #3396, opened 2025-03-17
Improve data source validation results

P4T12ICK:
This PR fixes all results from the newly added data source validation check.


ljstella (approved):
LGTM

PR #3399, opened 2025-03-17
Updating to manual test

patel-bhavin:
The risk message is dynamically generated in the SPL and it needs to be manually tested for integration testing.


ljstella (approved):
LGTM

PR #3423, opened 2025-03-26
Windows Service Created with Suspicious Service Name fix

0xC0FFEEEE:
### Details

This detection generates risk for all services regardless of lookup result, this is because the lookup provides a default value of `false` whereas the SPL seeks to exclude results with a null value.

Also tweaked the process_name eval to remove doublequotes.

### Checklist

- [ ] Validated SPL logic.


ljstella:
@0xC0FFEEEE Thanks for the PR!

We changed some of the typing around `default_match` in lookups recently, looks like the lookup definition still has a `default_match: false` setting- initially this was being typed as a boolean. When it was converted to be a string, which matches the transforms.conf.spec definition, this one may not have been updated. Have we tested this with default_match removed from the lookup? (I'm also reviewing the rest of our lookups that may have been affected by that and would like to make that change consistent across all of them)

0xC0FFEEEE:
Thanks for the context @ljstella

No I didn't test with an updated lookup definition as I assumed that the default_match option was intentional. I'll go ahead and remove it from the lookup and revert to the original `| where isnotnull(tool_name)` statement.

ljstella (approved):

patel-bhavin:
thank you @0xC0FFEEEE for the fix! 🚀

patel-bhavin (approved):

PR #3425, opened 2025-03-26
New - Detect Large ICMP Traffic

patel-bhavin:
Based on feedback from Rahul Haridas for - Detect Large Outbound ICMP Packets! 

Deprecated this detection in favor of a better named detection . The detection logic remains the same!


0xC0FFEEEE:
@patel-bhavin
The drilldown for this detection uses src and dest but the search itself uses src_ip and dest_ip

patel-bhavin:
@0xC0FFEEEE - I am not sure, it does look like the drilldowns are using correct fields 🤔 
image

ljstella (approved):
LGTM

PR #3430, opened 2025-03-27
bug in cisco detection

patel-bhavin:
remove extra tick!


patel-bhavin:
Testing in CI without manual_test flag

image

pyth0n1c (approved):
approved, this looks good!

PR #3437, opened 2025-04-01
fix weird attack data links

pyth0n1c:
These are a number of minor issues with attack_data links that should be fixed.
First, some of them refer to RAW, non-log files, such as .txt files.  Those files
should be converted to LOG files so that they are available in gitlfs.

Second, some of the files use a slightly different path, such as a path 
that includes /refs/, when they should not.

This is true for a handful of production detections as well as non-production detections, 
such as experimental content.

However, even if something is experimental or deprecated, if it has an attack_data link, that link
should be validated to be correct. Right now, those are missed because that validation only happens
for tested content at `contentctl test` runtime.

When possible, we will move this validation to `contentctl validate` time when a local copy of the 
attack_data repo is present.


detections/endpoint/active_directory_privilege_escalation_identified.yml
```diff
@@ -66,10 +66,4 @@ tags:
   - Splunk Enterprise Security
   - Splunk Cloud
   security_domain: endpoint
-tests:
-- name: True Positive Test
-  attack_data:
-  - data: 
-      https://raw.githubusercontent.com/splunk/attack_data/master/datasets/attack_techniques/T1484/privesc/priv_esc.log
```
pyth0n1c: I cannot find a fitting file for this in the current git repo or the git history.  At one point this detection pointed at the following file, which does exist, but it is not correct. The name of the Analytic Story the search references, `Active
  Directory Privilege Escalation`, does not appear in ANY raw content in the attack_data repo (including old history, as far as I can tell).
https://github.com/splunk/security_content/blame/62e859ba7b7407a9418d9b64f1869f579ffe8dd4/detections/endpoint/active_directory_privilege_escalation_identified.yml#L56

Since test data is optional for correlation searches, this test data has been removed.

ljstella:
Okay, we're still failing, but we're failing because the link we've updated to changes in https://github.com/splunk/attack_data/pull/975 and that hasn't landed yet.

Edit: Now its passing after that landed.

ljstella:
CI here times out after 6 hours so...

Full package unit test results, from AWS:
```
summary:
  mode: All
  enable_integration_testing: false
  success: false
  total_detections: 1764
  total_tested_detections: 1572
  total_pass: 1571
  total_fail: 1
  total_skipped: 192
  total_untested: 0
  total_production: 1622
  total_experimental: 136
  total_deprecated: 6
  total_manual: 38
  success_rate: 99.9%
 ```
 Only failure is a known issue at the moment (see #3311)

patel-bhavin (approved):

patel-bhavin:
LGTM!  :shipit:

patel-bhavin (approved):

PR #3440, opened 2025-04-01
integration fixes for endpoint detections

patel-bhavin:
fixes for a few detections that failed on integration tests


patel-bhavin:
Merging this for integration test!

PR #3482, opened 2025-04-22
updating mitre ids

patel-bhavin:
New CTI release - T1574.002 is moved into T1574.001

image


ljstella (approved):
~LGTM~

Got some version issues

PR #3486, opened 2025-04-22
Update attack data - Delete ShadowCopy With PowerShell

patel-bhavin:
the original dataset had a LOT of events : added a new data set file with a single event that works for this detection:

Error: 
image


patel-bhavin:
image

pyth0n1c (approved):
Approving this, but we should seek to better understand the underlying error around `max_rawsize_perchunkk`

PR #3498, opened 2025-04-28
Cisco Secure Firewall Detections - Intrusion Correlation and Activity

nasbench:
additional Cisco FTD analytic that leverage Intrusion Events


patel-bhavin:
@nasbench  Pushed fixes for drilldown and risk objects. This PR looks good to me! Known failures with app inspect as we are waiting for the new creds!

patel-bhavin (approved):

PR #3500, opened 2025-04-29
Automated Splunk TA Update 250

patel-bhavin:


ljstella:
Still the same issue as #3311 , doesn't look like that issue was addressed in the new release. The rest of the detections all seem to be passing. We should follow up and see if this is scheduled, or if we should look to do something different with this one failing detection.

patel-bhavin:
this detection did not work with Windows Security data so I am not sure what is a good solution just yet. Will need to figure something else out! Will merge this PR as the other detections are passing!

pyth0n1c (approved):
I am approving because the 1 unit testing failure is a known failure.
It is currently tracked as a known defect in a trouble ticket.
This TA update does not appear to introduce any regressions.

PR #3514, opened 2025-05-08
Add missing Processes.process="*setzerodata*"

patel-bhavin:
We missed this filter on the search -

 - Add missing Processes.process="*setzerodata*" as per the original version.
 - Also, bump contentctl to 5.6.0
 - Removed two detections per validation checks 
 - Adding requirements.txt only to build since we need this branch in `develop`  branch for other git actions to be able to use this (will update git actions in a new PR for other jobs)

### DIFF
--- a/.github/workflows/appinspect.yml
+++ b/.github/workflows/appinspect.yml
@@ -18,12 +18,8 @@ jobs:
           
       - name: Install Python Dependencies and ContentCTL and Atomic Red Team
         run: |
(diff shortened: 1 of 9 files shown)


pyth0n1c (approved):
Content change looks good and automated testing is passing.
Changes to workflow are part 1 or 2 parts - another PR will follow with additional changes.

Some detections have also been removed as part of our content lifecycle plan.

PR #3515, opened 2025-05-08
Update CI to use requirements.txt

patel-bhavin:
updates  to inspect, test and data-dependabot to use requirements.txt to install contentctl . This will need to be bumped every time there a new version contentctl is released!


pyth0n1c (approved):
Looks good, no more use of CONTENTCTL_VERSION variable.
We should remove this from environment variables in the project settings.

PR #3521, opened 2025-05-13
Automated Splunk TA Update 264

patel-bhavin:


pyth0n1c (approved):
Results of the bump look good.
We also changed the CICD workflow such that if unit testing fails to run at all, it now shows as a failure instead of green.

PR #3529, opened 2025-05-19
Fix Reported Issues

### LINKED ISSUE #3397: [BUG] Ping Sleep Batch Command
The SPL for "Ping Sleep Batch Command" does not detect the events provided in https://media.githubusercontent.com/media/splunk/attack_data/master/datasets/attack_techniques/T1497.003/ping_sleep/sysmon.log

The where command filters for process_name=ping.exe, which does not exist in an event with the additional process criteria. Removing the process_ping macro will detect the second event.

Image

### LINKED ISSUE #3398: [BUG] Windows Process Commandline Discovery
The SPL for **Windows Process Commandline Discovery** is too specific.
It will not match processes like **wmic  process get CommandLine,Name,ProcessId** or **wmic  process get Name,CommandLine**

Suggestion for the where clause:
**where \`process_wmic\` AND Processes.process="\* process \*" AND Processes.process= "\* get \*" AND Processes.process="\*commandline\*"**

### LINKED ISSUE #3415: [BUG] ESCU - Schtasks scheduling job on remote system - Rule
### **Describe the bug**
The current detection is looking for the presence of the /s flag in the process command line arguments for schtasks.exe, indicating that it's being executed remotely 
Line: 
`(Processes.process="*/create*" AND Processes.process="*/s*")`

The problem I'm seeing in some environments is that with the presence of the wildcard directly after the "/s", this is causing it to alert on legitimate scheduled task creation events when they're defined with their schedule "/sc".

Apologies if I've read this incorrectly, but there needs to be a space after that wild card so that it is correctly looking for the presence of the /s flag.  

### **Expected behavior**
`(Processes.process="*/create*" AND Processes.process="*/s *")`

This should correctly identify when the /s flag is defined for a remote system, as it should always be in the format of "/s <server1>@contoso.com

### **App Version:**
 - ESCU: [7.3.2]
 - Splunk Security Essentials: [3.7.0]

### LINKED ISSUE #3495: [BUG] "Windows MSIExec DLLRegisterServer" Rule Matches Incorrect Processes Without Space Before /y or -y
**Rule Name:** ESCU - Windows MSIExec DLLRegisterServer

**Problem:** 
The rule matches on "*/y*" and "*-y*" without checking for space before the option, causing false positives.

For example, legitimate binaries or file names containing "-y" are incorrectly detected, because the wildcard match does not differentiate between file names and command-line arguments.

**Current search fragment causing problem:**
```spl
Processes.process IN ("*/y*", "*-y*")

nasbench:
This PR fixes the following reported issues.

- Fix #3397
- Fix #3398
- Fix #3415
- Fix #3495

The fixes are straight forward as most of the issues have correctly pointed the bugs accurately. 

@tccontre can you please look at the fix for `Ping Sleep Batch Command`. I removed the macro as the analytic is looking for commands that are linked after a ping, and added a `&` and other variants of the strings.

### DIFF
--- a/detections/endpoint/ping_sleep_batch_command.yml
+++ b/detections/endpoint/ping_sleep_batch_command.yml
@@ -1,11 +1,12 @@
 name: Ping Sleep Batch Command
 id: ce058d6c-79f2-11ec-b476-acde48001122
-version: 9
-date: '2025-05-02'
+version: 10
+date: '2025-05-19'
 author: Teoderick Contreras, Splunk
 status: production
 type: Anomaly
-description: The following analytic identifies the execution of ping sleep batch commands.
+description: |
+  The following analytic identifies the execution of ping sleep batch commands.
   It leverages data from Endpoint Detection and Response (EDR) agents, focusing on
   process and parent process command-line details. This activity is significant as
   it indicates an attempt to delay malicious code execution, potentially evading detection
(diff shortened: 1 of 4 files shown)


tccontre:
LGTM buddy! thank you.

tccontre (approved):
LGTM

patel-bhavin (approved):

patel-bhavin:
Thanks both for getting these issues squashed ❤️ !

PR #3535, opened 2025-05-22
Update Network CIM Analytics With FTD Data

nasbench:
The following PR maps some of Network CIM analytics we have to the Cisco datasources and providing test data. Below is the full list of updated analytics:

- Detect Outbound LDAP Traffic
- Detect Outbound SMB Traffic
- Internal Horizontal Port Scan
- Internal Horizontal Port Scan NMAP Top 20
- Internal Vertical Port Scan
- Prohibited Network Traffic Allowed
- Protocol or Port Mismatch
- Protocols passing authentication in cleartext
- TOR Traffic

Due to the nature of the mapping with the CSC TA. The Analytic `Detect Remote Access Software Usage Traffic` cannot be used as is. The current config is mapping the `Application` field to the CIM `app` field. Which is technically accurate. But in order to use this detection as is, we need to map the `ClientApplication` field which does not have an equivalent in CIM.

For this a new equivalent analytic has been created `Cisco Secure Firewall - Remote Access Software Usage Traffic` along with a dedicated lookup that is directly taken from Cisco Secure Firewall App Detectors list.

### DIFF
--- a/detections/network/internal_horizontal_port_scan.yml
+++ b/detections/network/internal_horizontal_port_scan.yml
@@ -18,7 +18,7 @@ search: '| tstats `security_content_summariesonly` values(All_Traffic.action) as
   values(All_Traffic.src_category) as src_category values(All_Traffic.dest_zone) as
   dest_zone values(All_Traffic.src_zone) as src_zone values(All_Traffic.src_port)
   as src_port count from datamodel=Network_Traffic where All_Traffic.src_ip IN ("10.0.0.0/8","172.16.0.0/12","192.168.0.0/16")
-  by All_Traffic.src_ip All_Traffic.dest_port All_Traffic.dest_ip span=1s _time All_Traffic.transport All_Traffic.rule
+  by All_Traffic.src_ip All_Traffic.dest_port All_Traffic.dest_ip All_Traffic.transport All_Traffic.rule span=1s _time
   | `drop_dm_object_name("All_Traffic")` | eval gtime=_time | bin span=1h gtime |
   stats min(_time) as _time values(action) as action dc(dest_ip) as totalDestIPCount
   values(src_category) as src_category values(dest_zone) as dest_zone values(src_zone)
--- a/detections/network/internal_horizontal_port_scan.yml
+++ b/detections/network/internal_horizontal_port_scan.yml
@@ -1,12 +1,13 @@
 name: Internal Horizontal Port Scan
 id: 1ff9eb9a-7d72-4993-a55e-59a839e607f1
-version: 6
-date: '2025-05-02'
+version: 7
+date: '2025-05-22'
 author: Dean Luxton
 status: production
 type: TTP
 data_source:
 - AWS CloudWatchLogs VPCflow
+- Cisco Secure Firewall Threat Defense Connection Event
 description: This analytic identifies instances where an internal host has attempted
   to communicate with 250 or more destination IP addresses using the same port and
   protocol. Horizontal port scans from internal hosts can indicate reconnaissance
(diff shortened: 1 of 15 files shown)


detections/network/protocols_passing_authentication_in_cleartext.yml
```diff
@@ -13,22 +13,37 @@ description: The following analytic identifies the use of cleartext protocols th
...
+  search: '| from datamodel Risk.All_Risk | search normalized_risk_object IN ("$src_ip$")
+    starthoursago=168  | stats count min(_time) as firstTime max(_time) as lastTime
+    values(search_name) as "Search Name" values(risk_message) as "Risk Message" values(analyticstories)
+    as "Analytic Stories" values(annotations._all) as "Annotations" values(annotations.mitre_attack.mitre_tactic)
+    as "ATT&CK Tactics" by normalized_risk_object | `security_content_ctime(firstTime)`
+    | `security_content_ctime(lastTime)`'
+  earliest_offset: $info_min_time$
+  latest_offset: $info_max_time$
 rba:
   message: Potential Authentication in cleartext
```
pyth0n1c: I think we should probably include $dest_port$ here as it gives more information about the specific protocol (assuming default port) which is triggering the alert. I realize this change was not made as part of this PR.

detections/network/protocols_passing_authentication_in_cleartext.yml
```diff
@@ -13,22 +13,37 @@ description: The following analytic identifies the use of cleartext protocols th
   and other sensitive data to interception. If confirmed malicious, attackers could
   capture authentication details, leading to unauthorized access and potential data
   breaches.
-data_source: []
+data_source:
+- Cisco Secure Firewall Threat Defense Connection Event
 search: '| tstats `security_content_summariesonly` count min(_time) as firstTime max(_time)
```
pyth0n1c, line 18: Should we add an eval / IF or something similar (a lookup is probably too heavy) to map the specific ports in question to a service?
Or do we just assume users know what all of these services are?

detections/network/internal_vertical_port_scan.yml
pyth0n1c, line 1: I think src_ip should probably be a threat_object and dest_ip should probably be a risk_object.

detections/network/internal_horizontal_port_scan_nmap_top_20.yml
pyth0n1c, line 1: Should also likely have both threat (src) and risk (dest_zone, maybe, since we do not have explicit dest_ip?) objects

detections/network/internal_horizontal_port_scan.yml
pyth0n1c, line 1: Risk/threat objects again? Maybe since there is no dest_ip (presumably since there would be a huge number) it's actually the port that would be the risk_object?

pyth0n1c (changes requested):
No actual issues and all the tests + results were inspected and look good.
I do want to get some feedback, however, before approving and merging
with respect to some of the RBA fields, mostly around threat and risk objects.
Those fields were not actually updated as part of this PR, but since I took a 
look at them I want to make sure they are in line with our expectations.

detections/network/protocols_passing_authentication_in_cleartext.yml
nasbench: I changed the RBA message to `Allowed Traffic from $src_ip$ to $dest_ip$ over port $dest_port$. Which might indicate a potential authentication attempts over a  cleartext protocol.`. Which is better suited for the type of alert.

detections/network/protocols_passing_authentication_in_cleartext.yml
nasbench: As an improvement to the rule yes. This could be extended with a lookup containing a list of clear text protocols, default port, and some additional descriptions for better analysis.

I think this is better served as an update in the next release instead of this one. As this is release week.

pyth0n1c (approved):
After discussions with author and one last updated, approving this PR.
We will make further changes in the future, but these changes are a significant improvement for now.

PR #3536, opened 2025-05-22
Update  AWS Defense Evasion Impair Security Services to TTP

patel-bhavin:
Update  based on the slack thread in security content


nasbench (approved):

pyth0n1c (approved):
Changes and testing look good. Approved.

PR #3569, opened 2025-06-17
Correct "remote_access_software" Lookup to Match macOS Processes from NVM Input

gneff:
### Details

### Checklist

- [X] Validated SPL logic.
- [X] Confirm updates to lookups are handled properly.

Validated lookup changes with this query on the CSIRT Splunk Development cluster:
```
earliest=-24h index=nvm process_name=* AND dst_name!=*.cisco.com 
| stats values(dst_name) as Destination by logged_in_user_name, process_name 
| join process_name 
    [| inputlookup remote_access_software_combined
    | where isnotnull(process_name) 
    | table process_name, category, comment_reference, description, remote_appid, remote_domain ]
```
(The 'inputlookup' name is unique to the CSIRT Dev server and should normally reflect the standard remote_access_software name.)

- Updates to existing lookup files can be tricky, because of how Splunk handles application updates and the differences between existing lookup files being updated vs new lookups. You can read more [here](https://docs.splunk.com/Documentation/SplunkCloud/8.2.2203/Admin/PrivateApps#Manage_lookups_in_Splunk_Cloud_Platform) but the short version is that any changes to lookup files need to bump the date and version in the associated YAML file.


ljstella:
AppInspect failure in CI is expected for PRs coming from a fork.

lookups/remote_access_software.csv
```diff
@@ -1,4 +1,4 @@
-description,remote_domain,remote_utility,remote_utility_fileinfo,remote_appid,isutility,category,comment_reference,last_update
```
nasbench, line 1: This lookup is used by other analytics. Why not rename the field in your search instead?

gneff: The original query had a rename (left it in here commented out), but it never matched any processes on macOS hosts:
```
earliest=-24h index=nvm process_name=* AND dst_name!=*.cisco.com 
| stats values(dst_name) as Destination by logged_in_user_name, process_name 
| join process_name 
    [| inputlookup remote_access_software_combined
        ```| rename remote_utility as process_name``` 
    | where isnotnull(process_name) 
    | table process_name, category, comment_reference, description, remote_appid, remote_domain ]
```

nasbench: Here is a working example from the repo renaming the field to process for use https://github.com/splunk/security_content/blob/5fa73118ce9a3ca691f2f0d1bf5b3fbd3ee0d047/detections/endpoint/detect_remote_access_software_usage_process.yml#L23-L32

I'm no expert on SPL but I think this might be an issue in your search.

But imo the solution is not renaming the lookup to fit (only) your need.

patel-bhavin:
@gneff : i think @nasbench is correct here. This lookup file is used in several other detections and changing the lookup field here will cause many other detections to not work. 

I also think there may be an opportunity to make this detection work for Mac OS processes but we do not have appropriate attack data  to test that change. What data are you ingesting for processes from macOS hosts?

 Can you help provide a sample event and we can have this change tested ?

patel-bhavin:
@gneff : Any thoughts on the above ask wrt attack_data ?

patel-bhavin:
@gneff : hello there! Can you please help us with a sample event that  you have been testing with for Mac OS? That will help with testing this PR and getting it shipped!

nasbench:
I talked to Glen internally and he said that he will submit a new PR with changes. They figured the renaming part on their side and they would like to maintain the RMM list at this level.

Hence we can close this for now, until further notice from @gneff

PR #3571, opened 2025-06-18
Update print_processor_registry_autostart.yml

pyth0n1c:
risk message used a field where the datamodel prefix was still included. This meant that it failed the risk message substitution


nasbench (approved):

nasbench:
Approved unless there are more errors as suggested by eric

Issue #3566, opened 2025-06-17
[BUG] Kerberos Pre-Authentication Flag Disabled in UserAccountControl


ljstella:
### **Describe the bug**
This dataset is still using the multi-line Windows Event Logs

### **Expected behavior**
Regenerate dataset & update search to use XMLWinEventLog

### **App Version:**
 - ESCU: 5.7.0

PR #3578, opened 2025-06-24
XML Windows Event Log Cleanup

ljstella:
### Details

We've "declared victory" on using `XmlWinEventLog` instead of the classic multiline `WinEventLog` several times in the past, and we're still finding gaps. This PR will include all of the remaining cleanup, including altering the source macros to no longer reference the WinEventLog source or sourcetypes. 

### Checklist

- [ ] Validated SPL logic.

- Updates to existing lookup files can be tricky, because of how Splunk handles application updates and the differences between existing lookup files being updated vs new lookups. You can read more [here](https://docs.splunk.com/Documentation/SplunkCloud/8.2.2203/Admin/PrivateApps#Manage_lookups_in_Splunk_Cloud_Platform) but the short version is that any changes to lookup files need to bump the the date and version in the associated YAML file.


ljstella:
Experimenting with the detection in https://github.com/splunk/security_content/issues/3566, as its a known case of using multiline Windows Event Logs... and it appears that in the XML format, the manipulation of the "Don't Require Preauth" setting on an account does not produce an event with a field specific to that, but you'd have to determine that was the change from the difference in the OldUacValue and the NewUacValue. Which seems... _less than ideal_, as a single event gets created no matter how many changes are made at once, further complicating any math that would need to be done.

ljstella:
List of codes: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/b10cfda1-f24f-441b-8f43-80cb93e786ec

I seem to recall Splunk adding some bitwise functions and operators in the last few versions, evaluating if these could make this less painful.

ljstella:
Okay, re: #3566  

Looks like the `UserAccountControl` field contains a space-delimited list of configurations that were changed (well, sorta) and for no longer requiring preauth, we're looking for `%%2096` to be in the list.

ljstella:
Non-removed detections with `WinEventLog` as a sourcetype for at least one attack_data file:
- [ ] detections/endpoint/detect_excessive_account_lockouts_from_endpoint.yml
- [ ] detections/endpoint/detect_new_local_admin_account.yml
- [ ] detections/endpoint/disable_show_hidden_files.yml
- [ ] detections/endpoint/disabling_systemrestore_in_registry.yml
- [ ] detections/endpoint/print_spooler_adding_a_printer_driver.yml
- [ ] detections/endpoint/print_spooler_failed_to_load_a_plug_in.yml
- [ ] detections/endpoint/schedule_task_with_http_command_arguments.yml
- [ ] detections/endpoint/schedule_task_with_rundll32_command_trigger.yml
- [ ] detections/endpoint/short_lived_scheduled_task.yml
- [ ] detections/endpoint/short_lived_windows_accounts.yml
- [ ] detections/endpoint/unknown_process_using_the_kerberos_protocol.yml
- [ ] detections/endpoint/windows_rdpclient_connection_sequence_events.yml

ljstella:
Closing in favor of #3887

PR #3587, opened 2025-07-01
update deployments and baselines with modified scheduling

pyth0n1c:
Please use with the following PR, which also assigns throttling
https://github.com/splunk/contentctl/pull/422

This PR presently fails the appinspect check `check_for_gratuitous_cron_scheduling` - but this error should be ignored since it is being built with the production version of contentctl rather than the one in the link above.  Running it with the PR above, we see this check pass:
image


patel-bhavin:
Closing since this is an old PR and we have no near term plans of merging this feature. Lets re open this in a fresh branch when we want to introduce new scheduling configs

PR #3603, opened 2025-07-10
Updating sourcetype values to PascalCase

patel-bhavin:
Make this consistent with what is shipped by the TAs and better matching in the onboarding assistant

image


patel-bhavin:
Also no more lowercase "xmlwineventlog" string in the entire repo:

image

ljstella (approved):
LGTM

PR #3620, opened 2025-07-24
configure contentctl.yml to use attack data cache

tjgeorgen:
Add configuration to contentctl.yml for the splunk attack_data cache. This allows contentctl to make use of predownloaded attack_data logs and avoid intermittent issues with downloading during testing, as well as validate attack_data links when using the `--verbose` flag. The cache itself can be downloaded by following the instructions from contentctl found here: https://github.com/splunk/contentctl/blob/2b633b6207d028f40908b8d7a618544aaa3876ce/contentctl/objects/config.py#L283-L293


pyth0n1c (approved):
I believe this is correct.  I think that, since I cannot recall we have seen a time where a download failure occurs (when a file DOES exist) we do not need to add download of the cache to the content test workflow in GitHub.  As a point of reference, here is the output we would get when an attack_data file has the wrong link format, but the file exists:


https://github.com/splunk/security_content/actions/runs/16506831778/job/46679499186?pr=3620#step:5:69
image

Which _is_ what we would like to see.

PR #3630, opened 2025-08-04
Update detection name - Splunk AppDynamics Secure Application Alerts

patel-bhavin:
Updates to detection name to reflect the correct product - Splunk AppDynamics Secure Application Alerts

deprecate the old detection and to be removed in 5.14.0


patel-bhavin:
Testing with new macro:
image

MHaggis (approved):

PR #3640, opened 2025-08-11
Updating Title

patel-bhavin:
image

app.manifest:
image

Update the title to be in parity with what is on installed on Splunk.


pyth0n1c (approved):
I have tested impacted detections locally and they all look good!
image

PR #3671, opened 2025-09-08
Automated Splunk TA Update 382

patel-bhavin:

### DIFF
--- a/contentctl.yml
+++ b/contentctl.yml
@@ -29,229 +29,232 @@ mode: {}
 splunk_api_username: null
 post_test_behavior: pause_on_failure
 apps:
-- uid: 1621
-  title: Splunk_SA_CIM
-  appid: Splunk_SA_CIM
-  version: 6.0.4
-  description: description of app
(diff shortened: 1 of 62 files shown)


ljstella:
We've run into this before, and I forgot how it broke things like this 🤦 

attack_range's automation placed the tarball named `splunk-add-on-for-amazon-web-services-(aws)_7110.tgz` in the S3 bucket. Initially this caused CI here to fail because the path didn't match (we expected `splunk-add-on-for-amazon-web-services-aws_7110.tgz`) so it couldn't be found. After updating the config side here to look for `splunk-add-on-for-amazon-web-services-(aws)_7110.tgz`, we got a different error about Datamodels failing to set up. I don't remember if we previously investigated that bug in contentctl, but its likely buried deep in the testing code. 

Solution: Copying the file in the bucket so that it exists with the name without the parentheses as well. That lets testing actually run.

ljstella (approved):
Tests passing now :shipit:

PR #3688, opened 2025-09-18
issue_2980

### LINKED ISSUE #2980: Azure AD Multi-Source Failed Authentications Spike - Missing ADFSSignInLogs category
For signin activity from ADFS, category is ADFSSignInLogs. This correlation would miss those due to category=SignInLogs. 

I would recommend to either include ADFS, category IN (SignInLogs, ADFSSignInLogs) or maybe using a wildcard such as category=*SignInLogs (in case there are other type of  Signinlogs with similar events). 

Also for the filter, I think uniqueUserAgents = 1 should be removed(or changed to look for >=1). These strings can easily be scripted to change on the fly for every attempt and those instances would be missed with this constraint.  

Thanks

patel-bhavin:
Creates a fix for issue -  https://github.com/splunk/security_content/issues/2980


detections/cloud/azure_ad_multi_source_failed_authentications_spike.yml
```diff
@@ -1,8 +1,8 @@
 name: Azure AD Multi-Source Failed Authentications Spike
 id: 116e11a9-63ea-41eb-a66a-6a13bdc7d2c7
-version: 9
-date: '2025-05-02'
-author: Mauricio Velazco, Splunk
+version: 10
+date: '2025-09-17'
+author: Mauricio Velazco, Splunk, atgithub11, Github Community
```
nasbench: The issues highlighted a suggestion and does not qualify to be an author as this will claim this whole rule was written by a GitHub community contributor.

I suggest you credit them in the release notes (which is more than enough).

nasbench (approved):

PR #3689, opened 2025-09-22
fix_net_profile_analytic

tccontre:
### Details

### Checklist

- [ ] Validated SPL logic.


detections/endpoint/windows_set_network_profile_category_to_public_via_registry.yml
```diff
@@ -1,11 +1,11 @@
...
 data_source:
   - Sysmon EventID 13
 search: '| tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime FROM datamodel=Endpoint.Registry
```
nasbench, line 11: I assume the rule is looking for changes to a `Private` profile as described from the ref blog? If so then the value should be `0x00000001`.

detections/endpoint/windows_set_public_network_profile_via_registry.yml
```diff
@@ -1,11 +1,11 @@
-name: Windows Set Private Network Profile via Registry
+name: Windows Set Public Network Profile via Registry
```
nasbench: A better title would be

```suggestion
name: Windows Set Network Profile Category to Private via Registry
```

detections/endpoint/windows_set_network_profile_category_to_public_via_registry.yml
nasbench, line 11: Also the search query is not excluding changes done via built-in processes like svchost and whatnot? If this is expected then I suggest you highlight that in the FP section by listing the default processes that might affect this.

tccontre: hmm base on microsoft blog the malware modification made was zero but they say it is private. I think they have some typo on describing the ttp's 
Screenshot 2025-09-22 at 16 26 50

nasbench: Yes that a typo in the blog. The docs say 
0 = public
1 = private

nasbench: @tccontre just a quick ping so that we can get this PR merged.

nasbench (approved):

PR #3705, opened 2025-10-01
Fix Issues & Updates

### LINKED ISSUE #3605: [BUG] False Positive in Analytic: Detect Rundll32 Application Control Bypass - setupapi
The analytic Detect Rundll32 Application Control Bypass - setupapi generates false positives due to overly broad matching of any rundll32.exe execution with setupapi.dll. Specifically, the current SPL does not limit detection to invocations of the LaunchINFSection function, as described in the analytic documentation.

For example, legitimate executions such as:

rundll32.exe setupapi.dll,InstallHinfSection

are incorrectly flagged by the detection logic, even though they do not use the LaunchINFSection function and are benign in many enterprise environments.
Expected behavior

The rule should only detect cases where rundll32.exe is used with setupapi.dll or iesetupapi.dll and the LaunchINFSection function is explicitly called. This would align the SPL with the analytic’s intended behavior and reduce false positives.
Screenshots

N/A
App Version:

    ESCU: 4.6.0 (or latest available if not version-locked)

    Splunk Security Essentials: N/A

Additional context

Current behavior in SPL:

Processes.process=*setupapi*

This condition captures all uses of setupapi.dll, regardless of the function invoked.

Suggested SPL Fix:

| tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime 
  from datamodel=Endpoint.Processes 
  where `process_rundll32` 
    Processes.process IN ("*setupapi.dll*", "*iesetupapi.dll*") 
    AND Processes.process="*LaunchINFSection*" 
  by Processes.action Processes.dest Processes.original_file_name Processes.parent_process 
     Processes.parent_process_exec Processes.parent_process_guid Processes.parent_process_id 
     Processes.parent_process_name Processes.parent_process_path Processes.process 
     Processes.process_exec Processes.process_guid Processes.process_hash Processes.process_id 
     Processes.process_integrity_level Processes.process_name Processes.process_path 
     Processes.user Processes.user_id Processes.vendor_product 
| `drop_dm_object_name(Processes)` 
| `detect_rundll32_application_control_bypass___setupapi_filter`

Justification:
This refinement ensures the detection logic strictly targets only invocations using LaunchINFSection, consistent with the documented analytic behavior. It helps reduce noise and improve detection fidelity.

### LINKED ISSUE #3684: Rule enhancement: Add "user" field to rule: WinEvent Scheduled Task Created Within Public Path
**Is your feature request related to a problem? Please describe.**
I think it would be a good idea to add the "user" field to this rule https://research.splunk.com/endpoint/5d9c6eee-988c-11eb-8253-acde48001122/?query=public So that you can see which user created the scheduled task.

**Describe the solution you'd like**

Adding user in the SPL like this for example:

``wineventlog_security` EventCode=4698 TaskContent IN ("*\\users\\public\\*", "*\\programdata\\*", "*\\temp\\*", "*\\Windows\\Tasks\\*", "*\\appdata\\*", "*\\perflogs\\*") 
| stats count min(_time) as firstTime max(_time) as lastTime by Computer, TaskName, TaskContent, user 
|  rename Computer as dest 
| `winevent_scheduled_task_created_within_public_path_filter``

### LINKED ISSUE #3702: [BUG] Common Ransomware Extensions - Regex Matching Double File Extension
### **Describe the bug**
The regex for "Detection: Common Ransomware Extensions" will never match on entries in the lookup, ransomware_extensions_lookup, that contain "double" file extensions. That is, patterns with more than one segment after the first period. Currently this only affects the following:

Line 16: *.Where_my_files.txt, Apocalypse
Line 24: *.bart.zip, Bart

Consider implementing the following regex to match on these edge cases:
`| rex field=file_name "(?<file_extension>(\.[^\.]+{1,2})$" `

### **Expected behavior**
The expected behavior is for this detection to match any file extension within the associated lookup.

### LINKED ISSUE #3714: Minor error in `message` field reference for linux_service_started_or_enabled
**Is your feature request related to a problem? Please describe.**
The risk message for linux_service_started_or_enabled is malformed, as the reference to the `dest` field is missing a `$`

https://github.com/splunk/security_content/blob/64a821213ed137e63a523c4390f3369002fe27f1/detections/endpoint/linux_service_started_or_enabled.yml#L57

**Describe the solution you'd like**
Adding a single `$` at the end should fix the issue. Could also capitalize the message for consistency.

```
message: A commandline $process$ that may create or start a service on $dest$
```

nasbench:
This PR fixes the following issues

- Fix #3684
- Fix #3702
- Fix #3605
- Fix #3714

It also updates/adds these rules

- Deprecate `Detect Rundll32 Application Control Bypass - advpack`, `Detect Rundll32 Application Control Bypass - setupapi` and `Detect Rundll32 Application Control Bypass - syssetup` in favor of a unified rule `Windows Application Whitelisting Bypass Attempt via Rundll32` since they are all related. - cc @MHaggis 
- Deprecate `Windows Change Default File Association For No File Ext` and replace it by `Windows Change File Association Command To Notepad`. The reason is that the original rule was incorrectly looking file without extensions, but in reality the TTP is a TA after encrypting files with a certain extensions (in this case `.enc` they add a handler for that extensions to open all files with notepad and the ransom note). cc @t-contreras 

- Updates `Add or Set Windows Defender Exclusion` with additional flags, namely `ControlledFolderAccessAllowedApplications` and `AttackSurfaceReductionOnlyExclusions`
- Reduce the `Attempt To Add Certificate To Untrusted Store` analytic to an `Anomaly` because this is found to be common by some installers.
- Fix the regex in `Common Ransomware Extensions` to account for double extension files.
- Updated `Linux Java Spawning Shell` and `Windows Java Spawning Shells` by removing `apache` and `w3wp.exe` respectively as they are unrelated to those rules. I did create `Web or Application Server Spawning a Shell` as a generic rule covering both linux and windows instances with a lot more web server and app server names.
- Updated `USN Journal Deletion` in order to filter for the deletion keyword at the search level not after the results for better performance.
- Updated `Windows Archived Collected Data In TEMP Folder` to used specific folders instead of just `\\temp\\` to avoid accidental FPs. As well as reducing it to an Anomaly rule since I found multiple cases where a match was found.
- Updated `Windows AutoIt3 Execution` - Added an OFN field and removed the overlap of the string.
- Updated `Windows Certutil Root Certificate Addition` by focusing on specific paths in order to avoid FPs. Since installers were found doing this as per elastic rule and VT results.
- Added the `--output-dir` flag to the rule `Windows Curl Download to Suspicious Path`
- Updated `Windows Information Discovery Fsutil` with additional flags, namely the `volume` sub-command that allows the discovery of disk information and was used by a TA before.
- Added `pwsh.exe` as a possible value to `Windows Remote Management Execute Shell`
- Enhanced `Windows Renamed Powershell Execution` by adding OFN fields, `powershell_ise.exe` and split the logic to be more accurate.
- Split the string logic in `Windows Rundll32 Apply User Settings Changes` to be more generic and avoid easy bypass using spaces.
- Updated `Windows Scheduled Task Created Via XML` with additional flags and OFN field as well as updated the different metadata sections.
- Added new analytic `Windows Symlink Evaluation Change via Fsutil`

### DIFF
--- a/detections/deprecated/detect_rundll32_application_control_bypass___advpack.yml
+++ b/detections/deprecated/detect_rundll32_application_control_bypass___advpack.yml
@@ -1,31 +1,31 @@
-name: Detect Rundll32 Application Control Bypass - setupapi
-id: 61e7b44a-6088-4f26-b788-9a96ba13b37a
-version: 11
-date: '2025-05-02'
+name: Detect Rundll32 Application Control Bypass - advpack
+id: 4aefadfe-9abd-4bf8-b3fd-867e9ef95bf8
+version: 12
+date: '2025-10-06'
 author: Michael Haag, Splunk
-status: production
+status: deprecated
 type: TTP
 description: The following analytic detects the execution of rundll32.exe loading
-  setupapi.dll and iesetupapi.dll via the LaunchINFSection function. This behavior
-  is identified using Endpoint Detection and Response (EDR) telemetry, focusing on
-  process creation events and command-line arguments. This activity is significant
-  as it indicates a potential application control bypass, allowing an attacker to
-  execute arbitrary script code. If confirmed malicious, this technique could enable
-  code execution, privilege escalation, or persistence within the environment, posing
-  a severe threat to system integrity and security.
+  advpack.dll or ieadvpack.dll via the LaunchINFSection function. This method is identified
+  using Endpoint Detection and Response (EDR) telemetry, focusing on command-line
+  executions and process details. This activity is significant as it indicates a potential
+  application control bypass, allowing script code execution from a file. If confirmed
+  malicious, an attacker could execute arbitrary code, potentially leading to privilege
+  escalation, persistence, or further network compromise. Investigate script content,
+  network connections, and any spawned child processes for further context.
 data_source:
 - Sysmon EventID 1
 - Windows Event Log Security 4688
 - CrowdStrike ProcessRollup2
 search: '| tstats `security_content_summariesonly` count min(_time) as firstTime max(_time)
-  as lastTime from datamodel=Endpoint.Processes where `process_rundll32` Processes.process=*setupapi*
+  as lastTime from datamodel=Endpoint.Processes where `process_rundll32` Processes.process=*advpack*
   by Processes.action Processes.dest Processes.original_file_name Processes.parent_process
   Processes.parent_process_exec Processes.parent_process_guid Processes.parent_process_id
   Processes.parent_process_name Processes.parent_process_path Processes.process Processes.process_exec
   Processes.process_guid Processes.process_hash Processes.process_id Processes.process_integrity_level
   Processes.process_name Processes.process_path Processes.user Processes.user_id Processes.vendor_product
   | `drop_dm_object_name(Processes)` | `security_content_ctime(firstTime)` | `security_content_ctime(lastTime)`
-  | `detect_rundll32_application_control_bypass___setupapi_filter`'
+  | `detect_rundll32_application_control_bypass___advpack_filter`'
 how_to_implement: The detection is based on data that originates from Endpoint Detection
   and Response (EDR) agents. These agents are designed to provide security-related
   telemetry from the endpoints where the agent is installed. To implement this search,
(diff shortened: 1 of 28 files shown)


patel-bhavin (approved):

patel-bhavin:
Looking good @nasbench : I set the deprecated version to 5.18.0 to keep the removal consistent with even numbered release

PR #3729, opened 2025-10-17
data source improvements

P4T12ICK:
bug fixes to sourcetypes in linux auditd data sources

### DIFF
--- a/data_sources/linux_auditd_add_user.yml
+++ b/data_sources/linux_auditd_add_user.yml
@@ -12,7 +12,7 @@ mitre_components:
 - OS API Execution
 - Application Log Content
 source: auditd
-sourcetype: auditd
+sourcetype: linux:audit
 separator: type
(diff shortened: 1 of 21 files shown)


data_sources/linux_auditd_add_user.yml
```diff
@@ -12,7 +12,7 @@ mitre_components:
 - OS API Execution
 - Application Log Content
 source: auditd
-sourcetype: auditd
+sourcetype: linux:audit
```
nasbench, line 15: If this is the new sourcetype emitted by the TA then all of the rules referencing these data sources will have to see their attack data links updated to reflect that.

Also the macro `linux_auditd` have to be updated.

nasbench (changes requested):

P4T12ICK:
I think this was a mistake from me. Closing this PR.

PR #3731, opened 2025-10-17
HTTP Request Smuggling

RavenTait:
### Details

Contains content around HTTP request smuggling. 

Detections:
HTTP Suspicious Tool User Agent
HTTP Request to Reserved Name
HTTP Rapid POST with Mixed Status Codes
HTTP Possible Request Smuggling
HTTP Duplicated Header

Analytic Story:
HTTP Request Smuggling.

### DIFF
--- a/detections/web/http_duplicated_header.yml
+++ b/detections/web/http_duplicated_header.yml
@@ -10,13 +10,16 @@ description: Detects when a request has more than one of the same header. This i
   and back-end servers parse HTTP requests by using ambiguous or malformed headers to hide malicious 
   requests within legitimate ones. Attackers leverage duplicate headers, particularly Content-Length
   and Transfer-Encoding, to cause different servers in the chain to disagree on where one request
-  ends and another begins.
+  ends and another begins. RFC7230 states that a sender MUST NOT generate multiple header fields with the same field
+  name in a message unless either the entire field value for that header field is defined as a comma-separated 
+  list or the header field is a well-known exception.
 data_source:
 - Suricata
 search: '`suricata` http.request_headers{}.name="*"
   | rename dest_ip as dest
   | spath path=http.request_headers{}.name output=header_names
   | mvexpand header_names
+  | where lower(header_names) != "set-cookie"
   | stats count by _raw, header_names, src_ip, dest
   | where count > 1
   | stats values(header_names) as duplicate_headers by _raw, count, src_ip, dest
--- a/detections/web/http_duplicated_header.yml
+++ b/detections/web/http_duplicated_header.yml
@@ -1,0 +1,76 @@
+name: HTTP Duplicated Header
+id: 1606cc5b-fd5f-4865-9fe3-0ed1eaec2df6
+version: 1
+date: '2025-10-15'
+author: Raven Tait, Splunk
+status: production
+type: Anomaly
+description: Detects when a request has more than one of the same header. This is commonly used 
+  in request smuggling and other web based attacks. HTTP Request Smuggling exploits inconsistencies in how front-end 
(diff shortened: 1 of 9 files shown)


detections/web/http_duplicated_header.yml
```diff
@@ -0,0 +1,76 @@
...
+data_source:
+- Suricata
+search: '`suricata` http.request_headers{}.name="*"
+  | rename dest_ip as dest
+  | spath path=http.request_headers{}.name output=header_names
+  | mvexpand header_names
+  | stats count by _raw, header_names, src_ip, dest
+  | where count > 1
+  | stats values(header_names) as duplicate_headers by _raw, count, src_ip, dest
+  | `http_duplicated_header_filter`'
```
nasbench, line 26: The `Set-Cookie` header is a known FP as it is often sent multiple times to setup different cookies/expire them.

So perhaps we should take that into consideration and filter it out?

Also the RFC 7230 states clearly that duplicate headers are not "allowed" (technically). Maybe you can incorporate that into the description. So that users have a better understanding. 

Similar to how tenable did it. https://www.tenable.com/plugins/was/113333

detections/web/http_request_to_reserved_name.yml
```diff
@@ -0,0 +1,79 @@
+name: HTTP Request to Reserved Name
```
nasbench: Because this is targeting ISS. Maybe explicitly state in the title (as you did in the description). So that people do not have a false assumption that this is generic.

lookups/scripting_tools_user_agents.csv
```diff
@@ -0,0 +1,42 @@
...
+Generic Scanner,*scanner*
+Generic Scanner,*scan/*
+Test or POC,*test*
+Test or POC,*debug*
+Test or POC,*poc*
+Test or POC,*demo*
+Test or POC,*hack*
+Test or POC,*exploit*
+Test or POC,*injection*
+Empty User Agent, "-"
```
nasbench: Sigma offers multiple rules for susp UAs. Perhaps you can increase coverage by adding some.

- https://github.com/SigmaHQ/sigma/blob/master/rules/web/proxy_generic/proxy_hello_world_user_agent.yml
- https://github.com/SigmaHQ/sigma/blob/master/rules/web/proxy_generic/proxy_ua_apt.yml
- https://github.com/SigmaHQ/sigma/blob/master/rules/web/proxy_generic/proxy_ua_cryptominer.yml
- https://github.com/SigmaHQ/sigma/blob/master/rules/web/proxy_generic/proxy_ua_frameworks.yml
- https://github.com/SigmaHQ/sigma/blob/master/rules/web/proxy_generic/proxy_ua_hacktool.yml
- https://github.com/SigmaHQ/sigma/blob/master/rules/web/proxy_generic/proxy_ua_malware.yml
- https://github.com/SigmaHQ/sigma/blob/master/rules/web/proxy_generic/proxy_ua_susp.yml

detections/web/http_duplicated_header.yml
RavenTait: Added filter for set-cookie and expanded description with info from RFC 7230

detections/web/http_request_to_reserved_name.yml
RavenTait: Updated name to reflect that its for IIS

lookups/scripting_tools_user_agents.csv
RavenTait: Expanded the list with ones I could verify from the SigmaHQ rules

nasbench (approved):
LGTM

PR #3737, opened 2025-10-21
Fixing spelling mistake present in two detections

Shotscape:


Shotscape:
Corrected a minor spelling error to ensure the executable name (WmiPrvSE) in the description fields matches the detection logic / correct exe name.

patel-bhavin:
Hello @Shotscape : Thank you for the PR! unfortunately due to some product limitations we cannot directly rename a search and instead we would have to replace this detection with a new one  and stage the current detection for deprecation. 

We have created a duplicate PR of this fix : https://github.com/splunk/security_content/pull/3738
Thank you!

PR #3755, opened 2025-11-04
Ollama data source bug

patel-bhavin:
data source validation failures due to the splunkbase app not found. This fixes this job and data source dependabot should be fixed! 
Screenshot

image


ljstella (approved):
LGTM

PR #3763, opened 2025-11-06
Fix - Windows Kerberos Local Successful Logon

patel-bhavin:
adding fillnull as the latest TA does not alias fields process_path and process_name and its not present in the raw event

Manual test added for this issue: https://github.com/splunk/splunk-add-on-for-microsoft-windows/issues/563


ljstella (approved):
Temp fix until TA change :shipit:

PR #3779, opened 2025-11-12
switching source and sourcetype

patel-bhavin:
fixes the detection failure - https://github.com/splunk/splunk-add-on-for-microsoft-windows/issues/563


ljstella (approved):
LGTM

PR #3792, opened 2025-11-17
Appinspect on develop

patel-bhavin:
inspect fails when code gets merged into develop : https://github.com/splunk/security_content/actions/runs/19443579254/job/55632585675

### DIFF
--- a/.github/workflows/appinspect.yml
+++ b/.github/workflows/appinspect.yml
@@ -9,10 +9,14 @@ jobs:
   appinspect:
     runs-on: ubuntu-latest
     steps:
-      - name: Check out the repository code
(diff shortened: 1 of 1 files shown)


ljstella:
Think we could just remove this?
https://github.com/splunk/security_content/blob/bf49526dba0701b42abd60d7209eb2f7021b378a/.github/workflows/appinspect.yml#L5-L7

Not sure we ever really check the outcome of this job, not sure it's worth to keep running.

patel-bhavin:
Fair enough, we dont check it- updated the PR as per comment!

ljstella (approved):
LGTM

PR #3802, opened 2025-11-24
Internal horizontal port scan nmap iteration

DipsyTipsy:
### Details
Suggesting a few minor fixes to "Internal horizontal port scan nmap" in order to speed up, and enrich, the query. These changes made the query ~50% faster in our environments.

Changes:
- Moving the logic from the 2nd stats statement into tstats, speeding up the query quite a bit.
- Moving to using firstTime/lastTime so we have the scope of the scanning.
- Adding `num_ports_scanned` as a field in order to simplify tuning.
- Minor Syntax fix for the risk message

### Checklist

- [x] Validated SPL logic.

### DIFF
--- a/detections/network/internal_horizontal_port_scan_nmap_top_20.yml
+++ b/detections/network/internal_horizontal_port_scan_nmap_top_20.yml
@@ -1,7 +1,7 @@
 name: Internal Horizontal Port Scan NMAP Top 20
 id: 3141a041-4f57-4277-9faa-9305ca1f8e5b
-version: 7
-date: '2025-10-14'
+version: 8
+date: '2025-12-02'
 author: Dean Luxton
 status: production
 type: TTP
--- a/detections/network/internal_horizontal_port_scan_nmap_top_20.yml
+++ b/detections/network/internal_horizontal_port_scan_nmap_top_20.yml
@@ -14,21 +14,28 @@ description: This analytic identifies instances where an internal host has attem
   or scanning activities, potentially signaling malicious intent or misconfiguration.
   By monitoring network traffic logs, this detection helps detect and respond to such
   behavior promptly, enhancing network security and preventing potential threats.
-search: '| tstats `security_content_summariesonly` values(All_Traffic.action) as action
-  values(All_Traffic.src_category) as src_category values(All_Traffic.dest_zone) as
-  dest_zone values(All_Traffic.src_zone) as src_zone values(All_Traffic.src_port)
-  as src_port count from datamodel=Network_Traffic where All_Traffic.src_ip IN ("10.0.0.0/8","172.16.0.0/12","192.168.0.0/16")
-  AND All_Traffic.dest_port IN (21, 22, 23, 25, 53, 80, 110, 111, 135, 139, 143, 443,
-  445, 993, 995, 1723, 3306, 3389, 5900, 8080) by All_Traffic.src_ip All_Traffic.src
-  All_Traffic.dest_port All_Traffic.dest_ip All_Traffic.dest All_Traffic.transport All_Traffic.rule span=1s _time |
-  `drop_dm_object_name("All_Traffic")`  | eval gtime=_time  | bin span=1h gtime  |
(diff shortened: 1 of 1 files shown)


patel-bhavin:
Hi @DipsyTipsy : the new search logic certainly seems more straightforward and perhaps VERY advisable to not have spans twice. Although, I have an open question that if we are looking for scanning activity should we have span=5m for example? Did you consider this for your environment  or does span=1h works good? Mostly just curious! 
 
 Will get this shipped  this in the next release

New Search ------ Old Search screenshot:

image

patel-bhavin (dismissed):

DipsyTipsy:
> Hi @DipsyTipsy : the new search logic certainly seems more straightforward and perhaps VERY advisable to not have spans twice. Although, I have an open question that if we are looking for scanning activity should we have span=5m for example? Did you consider this for your environment or does span=1h works good? Mostly just curious!
> 

We've tried both and generally end up having to filter out a lot of noise either way. Think the desired span her depends a bit on the desired confidence and severity of the finding? Currently running this with span=1h meets the desired MTTD for this kind of activity, however we might make a lower span version for using with RBA.

Just noticed that `dest_ports` are set as the risk_object here, does that make sense? Can't really see how that object can be used for anything?

patel-bhavin:
Makes sense! Lets keep the span as 1h. Although I do have a suggested change for the rba : We can update the search to have `values(All_Traffic.dest_ip) as dest_ip` to the search logic and update the rba section to have "dest_ip" instead of `dest_ports` with a risk score of 10 or less. Let me know what you think and I can push this fix to the detection. 

```
| tstats `security_content_summariesonly` values(All_Traffic.action) as action values(All_Traffic.src_category) as src_category values(All_Traffic.dest_zone) as dest_zone values(All_Traffic.src_zone) as src_zone values(All_Traffic.src_port) as src_port dc(All_Traffic.dest_ip) as totalDestIPCount values(All_Traffic.dest_ip) as dest_ip count min(_time) as firstTime max(_time) as lastTime values(All_Traffic.rule) as rule from datamodel=Network_Traffic
  where All_Traffic.src_ip IN ("10.0.0.0/8","172.16.0.0/12","192.168.0.0/16") AND All_Traffic.dest_port IN (21, 22, 23, 25, 53, 80, 110, 111, 135, 139, 143, 443, 445, 993, 995, 1723, 3306, 3389, 5900, 8080)
  by All_Traffic.src_ip All_Traffic.dest_port All_Traffic.transport span=1h _time
  | `drop_dm_object_name("All_Traffic")`
  | where totalDestIPCount>=250
  | eval dest_port=transport + "/" + dest_port
  | stats min(firstTime) as firstTime max(lastTime) as lastTime values(action) as action sum(totalDestIPCount) as totalDestIPCount values(src_category) as src_category values(dest_ip) as dest_ip values(dest_port) as dest_port dc(dest_port) as num_ports_scanned values(dest_zone) as dest_zone values(src_zone) as src_zone values(rule) as rule by _time src_ip
  | fields - _time
  ```
  
  Screenshot from test data associated with this detection:
  
  
image

DipsyTipsy:
That makes the most sense, however that would generate at least 250 risk events per detection hit, a bit unsure if that is reasonable? The events themself could be useful in rba, but the volume might be a bit excessive?

Are there any guidelines on amount of risk events to generate per finding?

nasbench (approved):

patel-bhavin:
> That makes the most sense, however that would generate at least 250 risk events per detection hit, a bit unsure if that is reasonable? The events themself could be useful in rba, but the volume might be a bit excessive?
> 
> Are there any guidelines on amount of risk events to generate per finding?

Yes, Good point and thank you mucho for testing it in your env. 

i think its not reasonable to have 250 events for a single detection and after chatting with the team: we think that attributing risk to the single `src_ip` is perhaps the best and this adheres to the spirit of this detection to find compromised hosts scanning your network.  Will get this update shipped in ESCU 5.19.0 and please let us know if this is an incorrect solution for this detection.

patel-bhavin (approved):

PR #3803, opened 2025-11-24
Concept Shippable Response Plans

xqi-splunk:
### Details

This PR adds response templates to security_content repo:
1. `response_templates/*.json`: Response templates provided by the response templates team
2. `.github/workflows/response_templates/template_script.py`: Script provided by the response templates team, which provides functionality merged different versions response templates and generate manifest file of response templates.
3. `.github/workflows/response_templates/validate_response_templates.py`: Script to validate response templates using `Draft7Validator` with provided OpenAPI spec from the response templates team `.github/workflows/response_templates/mcopenapi_public.yaml`
4. `.github/workflows/build.yml`: Add upload of `dist/api`.
5. `.github/workflows/build-response-templates.yml`: Workflow to build merged response templates and validate response templates on push/PR.

### Checklist

- [ ] Validated SPL logic.

The update doesn't have impact to detections related. Below is the screenshot of build-response-templates workflow:

Screenshot 2025-12-02 at 1 46 17 PM


.github/workflows/response_templates/template_script.py
```diff
@@ -0,0 +1,105 @@
...
+        print(f"Error during merging files: {e}")
+        raise
+
+
+def _get_template_mapping(directory):
+    path = Path(directory)
+    if not path.exists() or not path.is_dir():
+        raise ValueError(f"The directory {directory} does not exist or is not a directory.")
+    
+    files = [f for f in path.iterdir() if f.is_file()]
```
pyth0n1c: iterdir() does not return files in any paritcular order. 
Later the subsequent for loop iterates over this list.  This means that the order of the response plans can change from run to run.  Do these HAVE TO BE in strictly increasing over, (v1, v2, v3 etc?) or is (v1, v10, v2 etc) acceptable?

response_templates/template_script.py
```diff
@@ -0,0 +1,105 @@
...
+    if not files:
+        raise ValueError(f"No files found in the directory {directory} to merge.")
+    
+    template_to_file_mapping = collections.defaultdict(list)
+
+    for file in files:
+        file_name_no_ext= file.name.replace(".json", "")
+        name_split = file_name_no_ext.rsplit("_v", 1)
+        if len(name_split) != 2:
+            print(f"Skipping file {file.name}: does not match expected pattern '<template_name>_v<version>'")
```
pyth0n1c: If we find a JSON file that does not meet the format in this folder, I think this should be an EXCEPTION which stops processing, not a skipped file with a printed INFO message.

.github/workflows/response_templates/template_script.py
```diff
@@ -0,0 +1,105 @@
...
+    return template_to_file_mapping
+
+def merge_files(directory, output_dir):
+    try:
+        template_mapping = _get_template_mapping(directory)
+        for template_name, template_list in template_mapping.items():
+            out_template_name = f"{template_name}.json"
+
+            templates = []
+            for _, file in template_list:
```
pyth0n1c, line 104: No checking to ensure that versions are no "missing" version. For instance, should a response plan containing versions 1, 2, and 4 be valid? Or does it REQUIRE 1, 2, 3, and 4?

response_templates/template_script.py
```diff
@@ -0,0 +1,105 @@
...
+            out_template_name = f"{template_name}.json"
+
+            templates = []
+            for _, file in template_list:
+                with open(file, 'r') as in_file:
+                    content = in_file.read()
+                    templates.append(json.loads(content))
+            
+            with open(Path(output_dir) / out_template_name, 'w') as out_file:
+                out_file.write(json.dumps(templates))
```
pyth0n1c: We have merged response templates checked into the repo, but then we are overwriting them in this workflow.
I think we INTEND merged_response_templates to be manually updated (by running the template_script.py by hand) and committing those changes back to the repo - is that correct?  
If that's the case, then in CI/CD we should be comparing what this script generates to what is already checked into the repo and, if it doesn't match, throwing an error.

For other things in the security_content repo, we don't version control the outputs (for example, we don't version control the DA-ESS_ContentUpdate app package, just the content that goes towards building it). Following that paradigm, should we be version controlling the merged_response_templates, or just the individual ones and putting the merged ones in S3?

.github/workflows/build.yml
```diff
@@ -29,6 +29,11 @@ jobs:
           contentctl build --enrichments --enforce_deprecation_mapping_requirement
           mkdir artifacts
           mv dist/DA-ESS-ContentUpdate-latest.tar.gz artifacts/
+          cd response_templates
+          python template_script.py -d . -o ./merged_response_templates -m
+          cd ..
+          mkdir -p dist/api/response_templates
+          cp response_templates/merged_response_templates/* dist/api/response_templates/
```
pyth0n1c: This is building merged response templates, but they are never actually validated.  This workflow is different (and separate) from the validate-response-templates.yml workflow.

.github/workflows/validate-response-templates.yml
```diff
@@ -0,0 +1,36 @@
...
+          architecture: 'x64'
+
+      - name: Install dependencies
+        run: |
+          pip install pyyaml jsonschema
+
+      - name: Validate response templates
+        run: |
+          cd response_templates
+          python validate_response_templates.py -s mcopenapi_public.yaml -d . -m merged_response_templates/manifest.json --merged-dir merged_response_templates
```
pyth0n1c: This validates what is in the merged_response_templates dir, but it does NOT validate that the different versions in the response_templates dir is consistent with the merged_response_templates.

response_templates/template_script.py
pyth0n1c: Personal preference - I would pretty print these with `json.dumps(templates, indent=2)` for readability. Thoughts?

.github/workflows/response_templates/validate_response_templates.py
```diff
@@ -0,0 +1,237 @@
...
+            return [resolve_refs(item, schemas_dict) for item in schema_obj]
+        else:
+            return schema_obj
+
+    resolved_schema = resolve_refs(target_schema, schemas)
+
+    # Add JSON Schema draft version
+    resolved_schema['$schema'] = 'http://json-schema.org/draft-07/schema#'
+
+    return resolved_schema
```
pyth0n1c, line 76: Since we have constructed a JSONSchema by this point, I wonder if it is valuable to actually write out the .json file we constructed?  Should we actually have 2 different tools here:

One which ingests the openapi.yml file and creates JSONSchema files?
Another which ingests JSONSchema files and validates the response_plans against them?

Not requesting a change, but we should consider discussing.

.github/workflows/response_templates/validate_response_templates.py
```diff
@@ -0,0 +1,237 @@
...
+    parser.add_argument(
+        '-d', '--directory',
+        type=str,
+        default='.',
+        help='Directory containing response template JSON files'
+    )
+    parser.add_argument(
+        '-s', '--schema',
+        type=str,
+        default='mcopenapi_public.yaml',
```
pyth0n1c: I have some concerns with the completeness of the openapi_schema.yml file. 

1. Does not annotate all the required fields is should. For instance, some of the schemas do not have any fields marked require at all and others have very few.
2. Certain things which may be important, such as the `"order:"` key, which is an integer and as I understand it must be _STRICTLY_ incrementing, CANNOT be enforced via openAPI or JSONSchema.
3. None of the schemas in the file forbid EXTRA keys. Is that correct/intended behavior? 

We should confirm with the response_plan team this is intended and sufficient.

pyth0n1c (changes requested):
I would not mix the python files and schema files in the same directory.  Inside of the response_templates directory, I would probably have:

**merged_response_templates/** - exactly as it is today
either in merged_response_templates/bin or **.github/workflows/response_templates/** - template_script.py, validate_response_templates.py, mcopenapi_public.yml (I would convert this to yml not yaml, for consistency with the rest of the security_content repo, even if what the team provided was yaml)
I make this recommendation as there is no other "code" in the security_content repo other than the code which lives in the .github folder for use during CI/CD

.github/workflows/response_templates/template_script.py
xqi-splunk: Good catch Eric, let me ask the response plan team to see if they would like to preserve a fixed order in the merged response templates (e.g. v1 -> v3 increasing order).

response_templates/template_script.py
xqi-splunk: Agree, after we moved python scripts and openAPI spec to `.github/workflows/response_templates/`, non JSON files shall raised EXCEPTION.

.github/workflows/response_templates/validate_response_templates.py
xqi-splunk: The `.json` file would be useful for debugging. Would it be better to add a debug option to output the .json file? We probably won't commit the .json file to the repo (if we did, then we could skip the read from openAPI spec and use the .json file directly).

.github/workflows/response_templates/validate_response_templates.py
xqi-splunk: Agree, very few fields are marked as required (even for the version field). Going to checkout with the response team to see if this relatively loose schema definition is intended.

response_templates/template_script.py
xqi-splunk: Agree!

.github/workflows/response_templates/validate_response_templates.py
pyth0n1c: I think adding a dump of the JsonSchemas as a DEBUG option would be quite nice. 
I also agree with you that, since the openapiSchema.yml we are getting from the team is the ground truth, we shouldn't mess around with committing the derived jsonSchemas to the repo.

response_templates/template_script.py
xqi-splunk: After syncing with the response templates team, we decided that the path forward is the merged_response_templates to be generated by CI/CD. No need for comparing script generation and files live in the repo.

For version control, I don't think we have discussed that part with the response team. Need to further check with them. For current info we got, just have the individual response templates and the putting the merged ones in scs endpoint following the ESCU release pipeline.

.github/workflows/build.yml
xqi-splunk: Same as above, after syncing with the response templates team we decided that the path forward is the merged_response_templates to be generated by CI/CD. 

For the validation, are we going to incorporate the validation in the build process as well or in a separate workflow? It seems like in security_content, build.yml only handles build and testing are done by separate workflow.

.github/workflows/validate-response-templates.yml
xqi-splunk: Thanks for catching the workflow discrepancy. The validation workflow will be updated. Since we have decided that merged_response_templates generated by runtime, this will be updated to generate files first and then validate.

response_templates/template_script.py
xqi-splunk: Got reply from Henry saying that "**No plans for version controlling the merged templates + manifest at the moment**".

.github/workflows/response_templates/template_script.py
xqi-splunk: According to Henry, this behavior is acceptable.

> It shouldn't require all versions. For example, if the content team uploads v1, v2, v3, v4 and v3 they find a bug or exploitation. They can pull v3 from the git repo and the merged templates would still be valid.

.github/workflows/response_templates/template_script.py
xqi-splunk: This is considered as a nice to have from the response plan team. I will make a small update to the `template_script.py` so that the version is in strictly increasing order.

xqi-splunk: Updated to sort version by increasing order. Also for manifest, updated to sort by name, and version by increasing order.

response_templates/template_script.py
xqi-splunk: Changed to raise exception.

.github/workflows/validate-response-templates.yml
xqi-splunk: Added merged templates generation

response_templates/template_script.py
xqi-splunk: Updated. Also updated the output for manifest.

.github/workflows/response_templates/validate_response_templates.py
xqi-splunk: Added `--debug` to dump json schema

xqi-splunk:
> I would not mix the python files and schema files in the same directory. Inside of the response_templates directory, I would probably have:
> 
> **merged_response_templates/** - exactly as it is today either in merged_response_templates/bin or **.github/workflows/response_templates/** - template_script.py, validate_response_templates.py, mcopenapi_public.yml (I would convert this to yml not yaml, for consistency with the rest of the security_content repo, even if what the team provided was yaml) I make this recommendation as there is no other "code" in the security_content repo other than the code which lives in the .github folder for use during CI/CD

Moved `template_script.py`, `validate_response_templates.py`, `mcopenapi_public.yml` to **.github/workflows/response_templates/** and updated `mcopenapi_public.yaml` to `mcopenapi_public.yml` for consistency.

.github/workflows/response_templates/template_script.py
pyth0n1c: Looks good!

.github/workflows/build.yml
pyth0n1c: I don't think we should do validation in a different workflow (see my new comment on validate-repsonse-templates.yml).  
In the build.yml workflow, it's actually doing the equivalent of both:
`contentctl validate` AND `contentctl build`.  
I don't think for response templates, we have an equivalent of `contentctl test` (and that isn't the requirement) - what you've built is much more aligned with `contentctl validate`


I am **quite** surprised that we are not uploading everything in the dist/api folder as an artifact. That _should_ have been something we were doing long before the changes you've contributed related to response templates (so this has been an issue with the build workflow for years).

Can you add an upload of the entire `dist/api` folder at the end of the workflow? That will make is easy to examine the results of the build, which is extra important since the manifest/combined response plans are no longer tracked as separate files in the repo.

.github/workflows/validate-response-templates.yml
pyth0n1c, line 1: Since this is now taking place in the build.yml workflow, I don't think we need it duplicated here. I THINK it's just extra complexity.
Thoughts?

response_templates/template_script.py
pyth0n1c: Minor feedback - rather than treating the file name as a string and splitting off the extension:
`file_name_no_ext= file.name.replace(".json", "")` I think it's a bit better to use the proper path function:
`file_name_no_ext= file.stem`
I will share some more feedback via chat - sorry should have mentioned this earlier.

.github/workflows/validate-response-templates.yml
pyth0n1c: I think this is probably made redundant by the changes to the build.yml workflow?

response_templates/template_script.py
pyth0n1c: Looks good!

.github/workflows/response_templates/validate_response_templates.py
pyth0n1c: Given discussions with the Response Template Team, I think your implementation aligns with their request.

pyth0n1c (dismissed):
A little more feedback around organization and fixing a pre-existing deficiency in the build.yml workflow.

.github/workflows/build.yml
xqi-splunk: Added validation after running merged response templates generation. Removed the `validate-response-templates.yml`. Also add an upload of the entire `dist/api` folder in the `store_artifacts`.

.github/workflows/validate-response-templates.yml
xqi-splunk: Agree, after the changes this workflow becomes redundant. Removed the workflow.

response_templates/template_script.py
xqi-splunk: Agree, this is more useful in when there are terrible naming as you mentioned like `hello.jsonResponse_v1.json`. Also added check for whether non `.json` files found.

.github/workflows/response_templates/mcopenapi_public.yml
```diff
@@ -0,0 +1,4484 @@
...
+              "total_time_taken": 2.0
+            }
+          ]
+        }
+    ResponseTemplateVersion:
+      type: object
+      description: Version information for a response template.
+      properties:
+        version:
+          type: number
```
pyth0n1c: Should be integer

.github/workflows/response_templates/mcopenapi_public.yml
```diff
@@ -0,0 +1,4484 @@
...
+          example: "d81ff75d-d9fe-4618-9752-e2840e5aa147"
+          x-splunk-soar-connector-gen-parameter:
+            contains: ["response template id"]
+        _key:
+          type: string
+          description: Internal use only.
+          example: "d81ff75d-d9fe-4618-9752-e2840e5aa147"
+        version:
+          description: The response plan version.
+          type: number
```
pyth0n1c: Should be integer

.github/workflows/response_templates/mcopenapi_public.yml
xqi-splunk: Updated the version sorting in `template_script.py` using `int()`, and raise ValueError when conversion failed. Tested locally:
Screenshot 2025-11-25 at 2 43 16 PM
Screenshot 2025-11-25 at 2 43 11 PM

.github/workflows/response_templates/mcopenapi_public.yml
xqi-splunk: Same as above.

.github/workflows/response_templates/mcopenapi_public.yml
xqi-splunk: Waiting for Henry update on schema

xqi-splunk: Updated the `mcopenapi_public.yml` provided by Henry with fixes in version fields.

response_templates/template_script.py
pyth0n1c: Looks great, thanks!

.github/workflows/response_templates/template_script.py
```diff
@@ -0,0 +1,116 @@
+import argparse
```
patel-bhavin, line 19: lets add a short commented description of what this script does, consider adding example command line to execute this locally!

.github/workflows/response_templates/validate_response_templates.py
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/env python3
```
patel-bhavin, line 1: lets add a short commented description of what this script does, consider adding example command line to execute this locally!

.github/workflows/response_templates/mcopenapi_public.yml
```diff
@@ -0,0 +1,4493 @@
+openapi: 3.0.1
```
patel-bhavin, line 6: can we also add a commented description of how this file is used in CI

.github/workflows/response_templates/template_script.py
xqi-splunk: Added with example command line usage!

.github/workflows/response_templates/validate_response_templates.py
xqi-splunk: Added with example command line usage as well!

.github/workflows/response_templates/mcopenapi_public.yml
xqi-splunk: Sure! Updated!

xqi-splunk:
Tested security_content gitlab fetching from this feature branch: https://cd.splunkdev.com/threat-research/security_content/-/pipelines/32509507

The `build_response_templates` job passed in the above link.

patel-bhavin (approved):

user17286439:
Hi! I was wondering if response plan validation should be integrated into the contenctl. Are there any plans for this on your roadmap? Thanks.

PR #3810, opened 2025-11-27
fix_cwd_path_detections

tccontre:
### Updated detection to fix the weakness of PATH record in terms of execution of process in different working directory using CWD record

        modified:   detections/endpoint/linux_auditd_doas_conf_file_creation.yml
        modified:   detections/endpoint/linux_auditd_possible_access_or_modification_of_sshd_config_file.yml
        modified:   detections/endpoint/linux_auditd_possible_access_to_sudoers_file.yml
        modified:   detections/endpoint/linux_auditd_possible_append_cronjob_entry_on_existing_cronjob_file.yml
        modified:   detections/endpoint/linux_auditd_preload_hijack_via_preload_file.yml
        modified:   detections/endpoint/linux_auditd_unix_shell_configuration_modification.yml
        modified:   detections/endpoint/linux_magic_sysrq_key_abuse.yml

### search detection:

Screenshot 2025-11-27 at 15 33 03

### Example Detected Event A:

detected event where there is CWD but the doas.conf was tried to access outside the "/etc" dir.

Screenshot 2025-11-27 at 15 33 42

detected event where there is CWD value and it has the directory where the actual event process happened in this case in "/etc" folder:

Screenshot 2025-11-27 at 15 34 23

detected event where there is no CWD generated 

Screenshot 2025-11-27 at 15 35 22

### msg value of event that access doas.conf outside /etc folder or without CWD

### Checklist

- [ ] Validated SPL logic.


nasbench:
In Review

nasbench (approved):

nasbench:
Updated the detections to include the following 

- Group by actual auditd ID instead of the full msg. For this added an extraction in `audit_id`
- Added `nametype` filter where needed and to the output fields.
- Improved descriptions and overall indentation as well as RBA message.
- Added a new DS for type CWD

Most of the detections titles / type have to be updated in the future (something to consider for future releases) as they are not accurate. Also because the detection only focuses on PATH type auditd events they are not super useful to give context. An additional future improvement would be to include related EXECVE or PROCTITLE in order to exactly assess the level of suspiciousness.

Because this would lead to more research and future changes. I decided to push for this to get merged so at least we avoid the bypass and think of future improvements.

@t-contreras please do create a ticket for this internally so that we can track future enhancements.

PR #3811, opened 2025-11-28
chore: user_agent field fix in both the Azure AD authentication detections. Minor spelling mistake fixes.

Shotscape:
### Details:

The CIM compliant user_agent field was not being parsed correctly as the `| rename properties.* as *` line renames the field before the `| rename properties.userAgent as user_agent` line is able to rename it.

Updated some minor issues with the risk message and known false positive fields stating that the `Azure AD High Number Of Failed Authentications For User` search bin's every 5 minutes.

Lastly, fixed the word minutes being present twice in the risk message of the `Azure AD High Number Of Failed Authentications From Ip` search.

### Checklist:

- [x] Validated SPL logic.

### DIFF
--- a/detections/cloud/azure_ad_high_number_of_failed_authentications_for_user.yml
+++ b/detections/cloud/azure_ad_high_number_of_failed_authentications_for_user.yml
@@ -15,7 +15,11 @@ description: The following analytic identifies an Azure AD account experiencing 
   based on their specific environment to reduce false positives.
 data_source:
 - Azure Active Directory
-search: '`azure_monitor_aad` category=SignInLogs properties.status.errorCode=50126 properties.authenticationDetails{}.succeeded=false 
+search: |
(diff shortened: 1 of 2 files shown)


nasbench:
The TA actually extracts user_agent from SignInLogs so no need for a rename anymore.

```
EVAL-user_agent = case(operationName IN ("Sign-in activity"),'properties.userAgent', \
                        true(),mvindex('properties.additionalDetails{}.value',mvfind('properties.additionalDetails{}.key',"^User-Agent$")))
```

You can see that even if a removed the rename the field is populated.

image

nasbench (approved):
Change LGTM

patel-bhavin (approved):

patel-bhavin:
inspect failures are expected from forks! Thank you for the fixes @Shotscape

PR #3816, opened 2025-12-04
Add tuoni content and named pipe detection

RavenTait:
### New Analytic Story:

- Tuoni

### New Analytics

- Windows PUA Named Pipe
- Windows RMM Named Pipe
- Windows Suspicious C2 Named Pipe
-  Windows Suspicious Named Pipe

### New Lookups

- pua_named_pipes
- suspicious_c2_named_pipes
- suspicious_named_pipes
- suspicious_rmm_named_pipes

### Breaking Changes

- Deprectated `Cobalt Strike Named Pipes` analytic as it's now included in this larger collection of pipe names.


nasbench (commented):
Overall LGTM. But I highly suggest we split the pipe csv into different categories.

Based on the type column you have, I can suggest C2, RMM, PUA (for stuff like PsExec, Dotnet)

That way it'll be easier to manage FPs and RBA / Findings would not blow up.

The C2 category can be TTP and the rest can be anomaly with a much lower risk score.

nasbench (approved):
LGTM

PR #3820, opened 2025-12-08
Automated Splunk TA Update 473

patel-bhavin:


nasbench:
Noticed that the TA-ollama is not part of contentctl.yml (the TA is not in ttack-range-appbinaries). Can we fix this?

cc @ljstella

ljstella:
Screenshot 2025-12-08 at 8 44 13 AM

Unclear why it was never added to `contentctl.yml` already, but each version has been getting put in the bucket.

nasbench (approved):

PR #3822, opened 2025-12-08
Add React2Shell Snort Mapping

nasbench:
This PR adds mapping content for the recently released snort detection related to React2Shell


ljstella (approved):
LGTM

PR #3827, opened 2025-12-10
sesameop

tccontre:
### tagged
        modified:   detections/endpoint/executables_or_script_creation_in_suspicious_path.yml
        modified:   detections/endpoint/executables_or_script_creation_in_temp_path.yml
        modified:   detections/endpoint/registry_keys_used_for_persistence.yml
        modified:   detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
        modified:   detections/endpoint/windows_process_execution_in_temp_dir.yml
        modified:   detections/endpoint/windows_suspicious_process_file_path.yml
        modified:   detections/endpoint/winevent_scheduled_task_created_within_public_path.yml
        modified:   detections/endpoint/winevent_windows_task_scheduler_event_action_started.yml
        modified:   detections/network/suspicious_process_dns_query_known_abuse_web_services.yml
        modified:   detections/endpoint/executables_or_script_creation_in_temp_path.yml
        modified:   detections/endpoint/windows_boot_or_logon_autostart_execution_in_startup_folder.yml
        

### new 

        new file:   detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
        
### updated
        modified:   detections/endpoint/windows_ai_platform_dns_query.yml      
### story
        new file:   stories/sesameop.yml
        new file:   stories/promptflux.yml

### Checklist

- [ ] Validated SPL logic.


detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,78 @@
...
+  | `security_content_ctime(lastTime)` 
+  | `windows_potential_appdomainmanager_hijack_artifacts_creation_filter`
+how_to_implement: To successfully implement this search, you need to be ingesting
+  logs with the file name, file path, and process_guid executions from your endpoints.
+  If you are using Sysmon, you can use the Add-on for Linux Sysmon from Splunkbase.
+known_false_positives: Administrator or network operator can create file in ~/.ssh
+  folders for automation purposes. Please update the filter macros to remove false
+  positives.
+references:
+- https://www.microsoft.com/en-us/security/blog/2025/11/03/sesameop-novel-backdoor-uses-openai-assistants-api-for-command-and-control/
```
nasbench, line 48: Please add additional references related to the technique.

https://attack.mitre.org/techniques/T1574/014/ - Is a good start and has more references to abuse in the wild

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,78 @@
...
+  | where file_count = 3 AND  match_count > 0
+  | `security_content_ctime(firstTime)`
+  | `security_content_ctime(lastTime)` 
+  | `windows_potential_appdomainmanager_hijack_artifacts_creation_filter`
+how_to_implement: To successfully implement this search, you need to be ingesting
+  logs with the file name, file path, and process_guid executions from your endpoints.
+  If you are using Sysmon, you can use the Add-on for Linux Sysmon from Splunkbase.
+known_false_positives: Administrator or network operator can create file in ~/.ssh
+  folders for automation purposes. Please update the filter macros to remove false
+  positives.
```
nasbench: This needs to be updated

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,78 @@
...
+    | `security_content_ctime(lastTime)`'
+  earliest_offset: $info_min_time$
+  latest_offset: $info_max_time$
+rba:
+  message: A file $files_found$ is created in $file_path$ on $dest$
+  risk_objects:
+  - field: dest
+    type: system
+    score: 30
+  threat_objects: []
```
nasbench: Add files name as a threat object

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,78 @@
...
+  min(firstTime) AS firstTime max(lastTime) AS lastTime
+  BY dest process_guid
+  | eval exe_files = mvfilter(match(files_found, "\.exe$") AND NOT match(files_found, "\.exe\.config$"))
+  | eval config_files = mvfilter(match(files_found, "\.exe\.config$"))
+  | eval exe_base_names = mvmap(exe_files, replace(exe_files, "\.exe$", ""))
+  | eval config_base_names = mvmap(config_files, replace(config_files, "\.exe\.config$", ""))
+  | eval matching_base_names = mvmap(exe_base_names, if(mvfind(config_base_names, VALUE) != -1, VALUE, null()))
+  | eval file_count = mvcount(files_found)
+  | eval match_count = mvcount(matching_base_names)
+  | where file_count = 3 AND  match_count > 0
```
nasbench: You are counting all files created by a process `values(file_name) AS files_found` and then matching only on 3.

This could lead to a break in logic if the process in question creates more than 3 files at a time.

Is this intentional?

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,78 @@
...
+author: Teoderick Contreras, Splunk
+status: production
+type: Anomaly
+description: The following analytic detects the creation of an .exe file along with its corresponding .exe.config and a .dll in the same directory, which is a common pattern indicative of potential AppDomain hijacking or CLR code injection attempts. This behavior may signal that a malicious actor is attempting to load a rogue assembly into a legitimate application's AppDomain, allowing code execution under the context of a trusted process.
+data_source:
+- Sysmon EventID 11
+search: |
+  | tstats `security_content_summariesonly` count min(_time) AS firstTime max(_time) AS lastTime from datamodel=Endpoint.Filesystem
+  where Filesystem.file_name IN ("*.exe", "*.exe.config", "*.dll") AND Filesystem.file_path IN
+    ("*\\windows\\fonts\\*", "*\\temp\\*", "*\\users\\public\\*", "*\\windows\\debug\\*","*\\Users\\Administrator\\Music\\*", "*\\Windows\\servicing\\*", "*\\Users\\Default\\*", "*Recycle.bin*", "*\\Windows\\Media\\*", "*\\Windows\\repair\\*", "*\\PerfLogs\\*", "*\\programdata\\*")
```
nasbench: ProgramData could be prone to FPs. In some of my test systems I already have `.exe` and corresponding `.exe.config` files created.

stories/sesameop.yml
```diff
@@ -0,0 +1,20 @@
+name: SesameOp
+id: 26b6c7c5-351b-489f-8053-da6cbaa74479
+version: 1
+date: '2025-12-10'
+author: Teoderick Contreras, Splunk
+status: production
+description: SesameOp is a newly discovered backdoor that abuses the OpenAI Assistants API as its command-and-control (C2) channel. Instead of using a traditional malicious server infrastructure, the malware loads a heavily obfuscated .NET DLL (Netapi64.dll / OpenAIAgent.Netapi64) which reaches out to the Assistants API to fetch encrypted, compressed commands and then executes them on the infected host. Results from these commands are likewise compressed, encrypted and sent back via the same legitimate API channel — effectively hiding malicious traffic in seemingly normal API calls. To evade detection, it injects into the host using .NET AppDomainManager injection, maintains persistence over time, and obfuscates communications via symmetric and asymmetric encryption plus compression.
```
nasbench: Do not use newly as this analytic story will live forever.
```suggestion
description: SesameOp is a backdoor that abuses the OpenAI Assistants API as its command-and-control (C2) channel. Instead of using a traditional malicious server infrastructure, the malware loads a heavily obfuscated .NET DLL (Netapi64.dll / OpenAIAgent.Netapi64) which reaches out to the Assistants API to fetch encrypted, compressed commands and then executes them on the infected host. Results from these commands are likewise compressed, encrypted and sent back via the same legitimate API channel, effectively hiding malicious traffic in seemingly normal API calls. To evade detection, it injects into the host using .NET AppDomainManager injection, maintains persistence over time, and obfuscates communications via symmetric and asymmetric encryption plus compression.
```

detections/endpoint/windows_ai_platform_dns_query.yml
```diff
@@ -1,27 +1,32 @@
...
+description: | 
+  The following analytic detects DNS queries initiated by the Windows AI Platform to domains associated with Hugging Face, a popular provider of machine learning models and services. Monitoring for such DNS requests is important because it can reveal when systems are reaching out to external AI platforms, which may indicate the use of third-party AI resources or the transfer of sensitive data outside the organization’s environment. Detecting these queries helps organizations enforce data governance policies, prevent unapproved use of external AI services, and maintain visibility into potential data exfiltration risks. Proactive monitoring ensures better control over AI model usage and organizational data flows.
```
nasbench: Change the description to be generic instead of only focusing on hugging face. Since now it also touches on OpenAI

detections/endpoint/windows_ai_platform_dns_query.yml
```diff
@@ -1,27 +1,32 @@
...
 data_source:
-- Sysmon EventID 22
-search: '`sysmon` EventCode=22 process_name IN ("python.exe", "cmd.exe", "rundll32.exe","powershell.exe", "pwsh.exe") QueryName= "router.huggingface.co"
+  - Sysmon EventID 22
+search: |
+  `sysmon` EventCode=22 (process_name IN ("python.exe", "cmd.exe", "rundll32.exe","powershell.exe", "pwsh.exe")  OR
+  Image IN ("*\\windows\\fonts\\*", "*\\windows\\temp\\*", "*\\users\\public\\*", "*\\windows\\debug\\*", "*\\Users\\Administrator\\Music\\*", "*\\Windows\\servicing\\*", "*\\Users\\Default\\*", "*Recycle.bin*", "*\\Windows\\Media\\*", "*\\Windows\\repair\\*", "*\\PerfLogs\\*")) AND
```
nasbench: Any particular reason this detection is not based on a blacklist approach?

Instead of focusing only on these paths and processes. Why not exclude web browsers and capture everything else?

I assume you want to know when these queries are initiated. Since you have a risk of 15, I think its acceptable.

tccontre: noted on that. thanks

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
tccontre: done

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
tccontre: done.

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
tccontre: done

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
tccontre: noted thanks for that one will removed programdata

macros/is_not_browser_app_macro.yml
```diff
@@ -0,0 +1,3 @@
+definition: lookup update=true browser_app_list browser_process_name AS process_name OUTPUT isAllowed | search isAllowed!=true
+description: This macro is designed to identify non-browser applications process name. Remove or filter as needed based.
```
nasbench: Why create a macro instead of using this directly in the detection? As we've discussed last time unnecessary macros should be avoided to make it clear to the end user what the rule is doing

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,90 @@
...
+  latest_offset: $info_max_time$
+rba:
+  message: A file $files_found$ is created in $file_path$ on $dest$
+  risk_objects:
+  - field: dest
+    type: system
+    score: 20
+  threat_objects: 
+  - field: file_name
+    type: file_name
```
nasbench, line 75:
```suggestion
  - field: file_name
    type: file_name
  - field: file_path
    type: file_path
```

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,90 @@
...
+  search: '| from datamodel Risk.All_Risk | search normalized_risk_object IN ("$dest$")
+    starthoursago=168  | stats count min(_time) as firstTime max(_time) as lastTime
+    values(search_name) as "Search Name" values(risk_message) as "Risk Message" values(analyticstories)
+    as "Analytic Stories" values(annotations._all) as "Annotations" values(annotations.mitre_attack.mitre_tactic)
+    as "ATT&CK Tactics" by normalized_risk_object | `security_content_ctime(firstTime)`
+    | `security_content_ctime(lastTime)`'
+  earliest_offset: $info_min_time$
+  latest_offset: $info_max_time$
+rba:
+  message: A file $files_found$ is created in $file_path$ on $dest$
```
nasbench: This message is not super helpful. What do you want to showcase here all the files or one of them?

Also keep in mind that in a multi value field, every entry will generate a risk. So in this case every time this rule triggers it will generate 3 risk events.

Is this something that you want?

detections/endpoint/windows_ai_platform_dns_query.yml
```diff
@@ -1,27 +1,31 @@
...
-  | stats count min(_time) as firstTime max(_time) as lastTime 
-  by answer answer_count dest process_exec process_guid process_name query query_count reply_code_id signature signature_id src user_id
-  vendor_product QueryName QueryResults QueryStatus 
-  | `security_content_ctime(firstTime)`
-  | `security_content_ctime(lastTime)`
-  | `windows_ai_platform_dns_query_filter`'
+  - Sysmon EventID 22
+search: |
+  `sysmon` EventCode=22 QueryName IN ("router.huggingface.co", "api.openai.com")
+    | `is_not_browser_app_macro`
```
nasbench: As discussed, please use the lookup directly instead of a macro

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,90 @@
...
+  | `windows_potential_appdomainmanager_hijack_artifacts_creation_filter`
+how_to_implement: To successfully implement this search you need to be ingesting information
+  on process that include the name of the process responsible for the changes from
+  your endpoints into the `Endpoint` datamodel in the `Filesystem` node. In addition,
+  confirm the latest CIM App 4.20 or higher is installed and the latest TA for the
+  endpoint product.
+known_false_positives: Administrator or network operator can create appdomain config file relative to the .NET application as part of installation. Filtering is needed.
+references:
+- https://www.microsoft.com/en-us/security/blog/2025/11/03/sesameop-novel-backdoor-uses-openai-assistants-api-for-command-and-control/
+- https://attack.mitre.org/techniques/T1574/014/
```
nasbench, line 49:
```suggestion
- https://attack.mitre.org/techniques/T1574/014/ 
- https://gist.github.com/djhohnstein/afb93a114b848e16facf0b98cd7cb57b
- https://www.scworld.com/brief/appdomain-manager-injection-exploited-for-cobalt-strike-beacon-delivery
- https://jp.security.ntt/insights_resources/tech_blog/appdomainmanager-injection-en/
```

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,90 @@
...
+  | where file_count >= 3 AND exe_present = 1 AND config_present = 1 AND dll_present = 1 AND exe_base_names = config_base_names
+  | `security_content_ctime(firstTime)`
+  | `security_content_ctime(lastTime)`
+  | `windows_potential_appdomainmanager_hijack_artifacts_creation_filter`
+how_to_implement: To successfully implement this search you need to be ingesting information
+  on process that include the name of the process responsible for the changes from
+  your endpoints into the `Endpoint` datamodel in the `Filesystem` node. In addition,
+  confirm the latest CIM App 4.20 or higher is installed and the latest TA for the
+  endpoint product.
+known_false_positives: Administrator or network operator can create appdomain config file relative to the .NET application as part of installation. Filtering is needed.
```
nasbench: The paths that you are using are highly suspicious and should not have that amount of FP. Except for `\\temp\\` which might be hit by `C:\Windows\Temp` or its `appdata` equivalent.

So the FP should be minimal except for the location i mentioned.

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,94 @@
...
+  latest_offset: $info_max_time$
+rba:
+  message: A file $file_name$ is created in $file_path$ on $dest$
+  risk_objects:
+  - field: dest
+    type: system
+    score: 20
+  threat_objects: 
+  - field: file_names
+    type: file_name
```
nasbench: @ljstella question for you. How do multi value threat objects show up?

detections/endpoint/windows_potential_appdomainmanager_hijack_artifacts_creation.yml
```diff
@@ -0,0 +1,94 @@
...
+  search: '| from datamodel Risk.All_Risk | search normalized_risk_object IN ("$dest$")
+    starthoursago=168  | stats count min(_time) as firstTime max(_time) as lastTime
+    values(search_name) as "Search Name" values(risk_message) as "Risk Message" values(analyticstories)
+    as "Analytic Stories" values(annotations._all) as "Annotations" values(annotations.mitre_attack.mitre_tactic)
+    as "ATT&CK Tactics" by normalized_risk_object | `security_content_ctime(firstTime)`
+    | `security_content_ctime(lastTime)`'
+  earliest_offset: $info_min_time$
+  latest_offset: $info_max_time$
+rba:
+  message: A file $file_name$ is created in $file_path$ on $dest$
```
nasbench: @ljstella same question here. If the condition is grouping the file name into MVs. What's best to use here? Should we perhaps use `file_names` and would it show correctly.

patel-bhavin: Nas, multi value fields do not have the best interpretation when it comes to risk_message. I enabled this detection as is on Endor and you can check out how the risk event looks using this query :

```
index=risk risk_object="ar-win-1" source="ESCU - Windows Potential AppDomainManager Hijack Artifacts Creation - Rule"

```

you can either use file_name or file_names and it would show up in with those field_names in the risk index but I tend to like using the CIM field name - file_name and file_path so that this generated risk  event has CIM complaint field names that can be leveraged downstream. 


For this detection, there are 3 files_names and 3 file_paths that get created as threat_objects as a part of a single risk event.

patel-bhavin: Image

patel-bhavin: when you agree with using CIM field names here, i can make this change and I think we should remove these multi value field names from the risk_message variable which would perhaps give the best UI experience.

nasbench: I agree on using CIM generally but I think there could be issues.

First file_name is not part of the output of the detection, and second which file_name entry will be used or will all entries matched by the original condition be listed?

patel-bhavin: @nasbench : here is the output of the updated search with CIM fields : file_name, file_path is present in the output and only the entries that satisfy the final where clause will be listed in the output. 


image

patel-bhavin:
Updated the detection yaml! @nasbench - we should be good to get this shipped

patel-bhavin (approved):

PR #3839, opened 2025-12-17
Yml validation cleanups

pyth0n1c:
A few issues resolved in this:

1. Baselines do not have filters.  If we want/need filter macros for baselines, then we can either create them by hand or emit them with contentctl.
2. A number of search did NOT include `Splunk Cloud` as one of their `products`.  I assume this omission was an error. Is that correct? Is there ever a reason that ESCU will NOT support all of the following?
products:
  - Splunk Enterprise
  - Splunk Enterprise Security
  - Splunk Cloud


nasbench (commented):
LGTM but you need to bump the versions before I approve.

And all ESCU content should support the 3 products, so it was an omission.

ljstella (approved):
:shipit:

PR #3846, opened 2025-12-19
minor_update_drv_detections

tccontre:
### NOTES

This PR updates existing Splunk Security Content detections related to suspicious driver installation and filesystem activity. The changes expand coverage by adding additional known Windows driver-related folder paths and removing overly restrictive filters that previously caused true malicious activity to be excluded. These updates are intended to reduce false negatives and improve detection fidelity for driver-based threats.

### Updated

        modified:   detections/endpoint/windows_file_without_extension_in_critical_folder.yml
        modified:   detections/endpoint/windows_service_create_kernel_mode_driver.yml
        modified:   detections/endpoint/windows_suspicious_driver_loaded_path.yml

### Checklist

- [ ] Validated SPL logic.


detections/endpoint/windows_service_create_kernel_mode_driver.yml
```diff
@@ -17,15 +17,19 @@ data_source:
 - Sysmon EventID 1
 - Windows Event Log Security 4688
 - CrowdStrike ProcessRollup2
-search: '| tstats `security_content_summariesonly` count min(_time) as firstTime max(_time)
-  as lastTime from datamodel=Endpoint.Processes where Processes.process_name=sc.exe
-  Processes.process="*kernel*" by Processes.action Processes.dest Processes.original_file_name
+search: |
+  | tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime from datamodel=Endpoint.Processes 
+  where Processes.process_name=sc.exe (Processes.process="*kernel*" OR Processes.process="*filesys*")
```
nasbench:
```suggestion
    where (Processes.process_name=sc.exe OR Processes.original_file_name=sc.exe)  Processes.process IN ("*kernel*", "*filesys*") Processes.process="*type*"
```

detections/endpoint/windows_service_create_kernel_mode_driver.yml
```diff
@@ -39,6 +43,7 @@ known_false_positives: False positives may be present based on common applicatio
   adding new drivers, however, filter as needed.
 references:
 - https://www.aon.com/cyber-solutions/aon_cyber_labs/yours-truly-signed-av-driver-weaponizing-an-antivirus-driver/
+- https://whiteknightlabs.com/2025/11/25/discreet-driver-loading-in-windows/
```
nasbench, line 46:
```suggestion
- https://whiteknightlabs.com/2025/11/25/discreet-driver-loading-in-windows/
- https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/sc-config
```

detections/endpoint/windows_file_without_extension_in_critical_folder.yml
```diff
@@ -14,14 +14,19 @@ description: The following analytic detects the creation of files without extens
   in potential data loss and system inoperability.
 data_source:
 - Sysmon EventID 11
-search: '| tstats `security_content_summariesonly` count FROM datamodel=Endpoint.Filesystem
-  where Filesystem.file_path IN ("*\\System32\\drivers\\*", "*\\syswow64\\drivers\\*")
+search: |
+  | tstats `security_content_summariesonly` count FROM datamodel=Endpoint.Filesystem
+  where Filesystem.file_path IN("*\\WINDOWS\\inf","*\\WINDOWS\\System32\\drivers\\*", "*\\WINDOWS\\System32\\DriverStore\\FileRepository\\*","*:\Windows\\WinSxS\\*","*\\ProgramData\\Microsoft\\Windows Defender\\Definition Updates\\*")
```
nasbench:
```suggestion
  where Filesystem.file_path IN("*:\\Windows\\inf", "*:\\Windows\\System32\\drivers\\*", "*:\\Windows\\System32\\DriverStore\\FileRepository\\*", "*:\\Windows\\WinSxS\\*", "*:\\ProgramData\\Microsoft\\Windows Defender\\Definition Updates\\*")
```

detections/endpoint/windows_suspicious_driver_loaded_path.yml
```diff
@@ -15,12 +15,12 @@ description: The following analytic detects the loading of drivers from
...
 data_source:
 - Sysmon EventID 6
-search: '`sysmon` EventCode=6 ImageLoaded = "*.sys" NOT (ImageLoaded IN("*\\WINDOWS\\inf","*\\WINDOWS\\System32\\drivers\\*",
-  "*\\WINDOWS\\System32\\DriverStore\\FileRepository\\*","*:\Windows\\WinSxS\\*","*\\ProgramData\\Microsoft\\Windows
-  Defender\\Definition Updates\\*")) |  stats  min(_time) as firstTime max(_time)
-  as lastTime count by ImageLoaded dest dvc process_hash process_path signature signature_id
-  user_id vendor_product | `security_content_ctime(firstTime)` | `security_content_ctime(lastTime)`
-  | `windows_suspicious_driver_loaded_path_filter`'
+search: '`sysmon` EventCode=6 
+   NOT (ImageLoaded IN("*\\WINDOWS\\inf","*\\WINDOWS\\System32\\drivers\\*", "*\\WINDOWS\\System32\\DriverStore\\FileRepository\\*","*:\Windows\\WinSxS\\*","*\\ProgramData\\Microsoft\\Windows Defender\\Definition Updates\\*"))
```
nasbench: This needs some fixing

- `\\WINDOWS\\inf` is missing a wildcard
- Sometimes `WINDOWS` is reported as `SystemRoot` so i suggest you add that option for all.
- `ProgramData`, `Program Files` and `Program Files (x86)` are also legit places that could load a driver.

Overall I suggest you make the list a whitelist instead of a blacklist or else we would have to reduce this to hunting rule

detections/endpoint/windows_file_without_extension_in_critical_folder.yml
nasbench: Can you provide some reference on why did you add these specific paths? Also why only these?

detections/endpoint/windows_suspicious_driver_loaded_path.yml
tccontre: as suggested I will change the list path to known suspicious path including temp, download and appdata

detections/endpoint/windows_file_without_extension_in_critical_folder.yml
tccontre: There is no additional reference driving this modification. The change simply expands coverage of known driver directory paths that may be abused by this technique, similar to how HermeticWiper leveraged its driver component. Based on the TTPs associated with these driver paths and detection logic, I didn't expect that this update will introduce unintended behavior. However, I can revert the change to the original logic if you think this expansion of coverage is not needed.

tccontre: I reverted back the detection logic. just added the reference of our blog to it

nasbench: Why did you revert the logic? I did not say anything about unintended behavior. I am just looking for the reasoning behind the paths.

The original logic was using what `hermeticwiper` used but I think this can be extended to things beyond drivers. Adding other folders will not hurt but its best to explain the reasoning and enhance the description.

tccontre: I just reverted back the detection as safe mode, while looking for possible reference that I can use while collecting more paths that I can use to expand this detection. I agree the no extension logic checking itself  of this detection can be a strong indication of suspicious events in most user folder in Windows OS. even in program files I think except for software that may use compressed file without file extension which is very un-ussual. I'll update it later while compiling the known folder path in drivers first and might update the description.
tnx

tccontre: now I pushed the updated one with more folder path of known drivers path. I hope I was able to cover the known one for this one. still looking for possible coverage of this even in non driver file type.

patel-bhavin (dismissed):

detections/endpoint/windows_file_without_extension_in_critical_folder.yml
```diff
@@ -1,33 +1,34 @@
...
 data_source:
 - Sysmon EventID 11
-search: '| tstats `security_content_summariesonly` count FROM datamodel=Endpoint.Filesystem
-  where Filesystem.file_path IN ("*\\System32\\drivers\\*", "*\\syswow64\\drivers\\*")
+search: |
+  | tstats `security_content_summariesonly` count FROM datamodel=Endpoint.Filesystem
+  where Filesystem.file_path IN("*\\System32\\drivers\\*", "*\\syswow64\\drivers\\*", "*\\WINDOWS\\inf\\*","*\\Program Files*", "*\\WINDOWS\\System32\\DriverStore\\*","*:\\Windows\\WinSxS\\*","*\\ProgramData\\Microsoft\\Windows Defender\\Definition Updates\\*","*systemroot*","*\\ProgramData\\Microsoft\\Windows Defender\\Platform\\*","*\\Windows\\servicing\\*", "*\\Windows\\ELAMBKUP\\*","*\\Windows\\Boot\\*","*\\System32\\Boot\\*","*\\System32\\Recovery\\*", "C:\\AMD\\*", "C:\\OEM\\*")
```
nasbench:
```suggestion
  where Filesystem.file_path IN("*\\System32\\drivers\\*", "*\\syswow64\\drivers\\*", "*\\WINDOWS\\inf\\*","*\\Program Files*", "*\\WINDOWS\\System32\\DriverStore\\*","*:\\Windows\\WinSxS\\*","*\\ProgramData\\Microsoft\\Windows Defender\\Definition Updates\\*","*\\ProgramData\\Microsoft\\Windows Defender\\Platform\\*","*\\Windows\\servicing\\*", "*\\Windows\\ELAMBKUP\\*","*\\Windows\\Boot\\*","*\\System32\\Boot\\*","*\\System32\\Recovery\\*", "C:\\AMD\\*", "C:\\OEM\\*")
```

nasbench (approved):

detections/endpoint/windows_file_without_extension_in_critical_folder.yml
nasbench: Removed this as it could introduce FPs for those using the env variable. `%systemroot%`, which resolves to `C:\Windows`

PR #3847, opened 2025-12-19
Move Unused Macros to Deprecate Sub-Folder

nasbench:
This PR creates a new sub-folder in the macros folder to list macros no longer in use (AKA deprecated).


patel-bhavin (approved):

patel-bhavin:
Looking good!

Issue #3809, opened 2025-11-27
Improving Performance of Detection Rules That Are Using Broad tstats Filters


a5hk:
### Summary

While investigating performance issues on our Splunk environment, I identified several detection rules from the **ES Content Updates** project that consistently consume excessive resources. The performance impact appears to stem from search design patterns that can be optimized. This issue summarizes the problematic patterns and lists specific rules affected.

### Observed Performance Problems
**1. Inefficient `tstats` Usage (Wide Filters + Large `BY` Clauses)**

Many rules use a `tstats` search with:

- A very broad filter (e.g., `Processes.file_name!="unknown"`)
- A `BY` clause containing too many fields
- A follow-up `lookup`, `join`, or secondary search to narrow the results after the fact

This leads to:

- Large result sets being produced unnecessarily
- High memory and CPU usage
- Delays in downstream searches (lookups, joins, subsearches)

**Additional Impact**

Including too many fields in the `BY` clause also causes result loss:
- Several of these fields are not required or not recommended by the CIM data model.
- These fields may be null or missing depending on the source (e.g., _Windows Event Log 4688_).
- Null values in several `BY` fields cause distinct combinations to collapse, which removes rows unintentionally.

**Example:**
In the `Suspicious Reg exe Process` rule:

- The subsearch groups results by `Processes.process_name` but discards it at the end using table, making the grouping unnecessary.
- The outer search includes fields like `Processes.parent_process_guid` in the `BY` clause, even though it is not CIM-recommended and not present in key data sources.

This pattern appears in several rules and results in unnecessary processing and inconsistent output.

**2. Duplicate or Near-Duplicate Rules**

Some rules could potentially be merged under a single analytic, as they perform very similar detections.

**Examples:**

- System User Discovery With whoami
- Windows System User Discovery Via quser

These could potentially be consolidated to reduce maintenance overhead and search executions.

### Rules Identified as High Resource Consumers

Below is the list of rules that, in our environment, consistently show high resource usage (based on job inspector metrics and internal monitoring):

- Windows LOLBAS Executed Outside Expected Path
- Single Letter Process On Endpoint
- System Processes Run From Unexpected Locations
- Windows Defacement Modify Transcodedwallpaper File
- Windows DotNet Binary in Non Standard Path
- Windows Command and Scripting Interpreter Hunting Path Traversal
- Windows NirSoft Utilities
- Common Ransomware Extensions
- Common Ransomware Notes
- Detect Remote Access Software Usage File
[...]

PR #3850, opened 2025-12-22
Analytic Enhancements and Fixes 

nasbench:
This PR is an initial effort to address some of the problematic rules referenced in #3809. The TL;DR is that some rules are relying on filtering later which hinders their performance a lot. Some are by design due to usage of lookups or features such as regex that are not available in tstats for example.

From the original list i have fixed what is possible and still working on the some of the rest.

### Updated Analytics [8]
- `Suspicious Email Attachment Extensions` - Added additional output fields in the `by` clause, as well as updated the RBA message.
- `DLLHost with no Command Line Arguments with Network` - Added a process filter in the process events search to reduce the sample size, instead of filtering via regex later.
- `Rundll32 with no Command Line Arguments with Network` - 
- `SearchProtocolHost with no Command Line with Network` - 
- `Single Letter Process On Endpoint` - Removed the usage of the regex instead opted for enumerating all possible values in the tstats. This will improve performance and will not lose any coverage.
- `System Processes Run From Unexpected Locations` - Added an additional filter location.
- `Windows Command and Scripting Interpreter Hunting Path Traversal` - Added an additional filter in the search, since it does not make sense to look for all processes and then filter them.
- `Windows LOLBAS Executed Outside Expected Path` - Added `WinSxS` as a new filter.

### Lookup Updates [3]
- `lookups/is_suspicious_file_extension_lookup.csv` - Since this lookup is only used in the `Suspicious Email Attachment Extensions` analytic. I updated it and reduced the search space to only thing that actually matter based on collected stats from our telemetry.
- `lookups/is_windows_system_file.csv` - Removed a lot of unnecessary filenames and reduced the search space.
- `lookups/lolbas_file_path.csv`


detections/endpoint/dllhost_with_no_command_line_arguments_with_network.yml
```diff
@@ -1,53 +1,61 @@
...
-  Processes.user Processes.user_id Processes.vendor_product
-  | `drop_dm_object_name(Processes)` 
+  as lastTime FROM datamodel=Endpoint.Processes where
+  (
+    Processes.process_name=dllhost.exe
+    OR
+    Processes.original_file_name=dllhost.exe
+  )
+  Processes.process IN (
+    "*dllhost",
```
patel-bhavin, line 28: i think these are missing * at the end. "*dllhost*"

detections/endpoint/rundll32_with_no_command_line_arguments_with_network.yml
```diff
@@ -15,24 +15,38 @@ description: The following analytic detects the execution of rundll32.exe withou
...
 data_source:
 - Sysmon EventID 1 AND Sysmon EventID 3
-search: '| tstats `security_content_summariesonly` count min(_time) as firstTime max(_time)
-  as lastTime FROM datamodel=Endpoint.Processes where `process_rundll32` AND Processes.action!="blocked"
+search: |
+  | tstats `security_content_summariesonly` count min(_time) as firstTime max(_time)
+  as lastTime FROM datamodel=Endpoint.Processes where
+  `process_rundll32`
+  Processes.process IN (
+    "*rundll32",
```
patel-bhavin, line 23: same as above comment

detections/endpoint/dllhost_with_no_command_line_arguments_with_network.yml
nasbench: They are not, these are mimicking a no CLI option hence no need for a wildcard

patel-bhavin (approved):

PR #3857, opened 2026-01-09
Add Snort/IOS Correlation and Other Things

nasbench:
This PR introduces a couple new analytics related to Snort/ Cisco IOS and updates to output and RBA fields of old snort based detections.

### New Analytics [5]
- Cisco Privileged Account Creation with HTTP Command Execution
- Cisco Privileged Account Creation with Suspicious SSH Activity
- Cisco Secure Firewall - Privileged Command Execution via HTTP
- Cisco Secure Firewall - SSH Connection to Non-Standard Port
- Cisco Secure Firewall - SSH Connection to sshd_operns

### Updated Analytics [29]
The updates mainly focused on changing the `dest_ip` and `src_ip` to `dest` and `src` respectively for ES compliance.

### Updated Data Sources [3]
Updated the output fields from `src_ip` to `src` for ES compliance.
- data_sources/cisco_secure_firewall_threat_defense_connection_event.yml
- data_sources/cisco_secure_firewall_threat_defense_file_event.yml
- data_sources/cisco_secure_firewall_threat_defense_intrusion_event.yml

### DIFF
--- a/data_sources/cisco_secure_firewall_threat_defense_connection_event.yml
+++ b/data_sources/cisco_secure_firewall_threat_defense_connection_event.yml
@@ -109,7 +109,7 @@ fields:
 - vendor_product
 - WebApplication
 output_fields:
-- src_ip
+- src
 - dest
 - dest_port
 - transport
--- a/data_sources/cisco_secure_firewall_threat_defense_file_event.yml
+++ b/data_sources/cisco_secure_firewall_threat_defense_file_event.yml
@@ -84,7 +84,7 @@ fields:
 - vendor_product
 - WebApplication
 output_fields:
-- src_ip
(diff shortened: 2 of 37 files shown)


patel-bhavin:
Largely looks okay, however looking for single snort IDs do not seem like scalable way of writing these and we should probably think of a lookup based solution.

patel-bhavin (approved):

nasbench:
> Largely looks okay, however looking for single snort IDs do not seem like scalable way of writing these and we should probably think of a lookup based solution.

It is not indeed. The reason these exist in this form are for a specific collab we are doing for the webinar. Overall, lookups are the way to go.

PR #3865, opened 2026-01-15
Refactor search query for prohibited network traffic

### LINKED ISSUE #3863: [BUG] ESCU - Prohibited Network Traffic Allowed - Rule
### **Describe the bug**
When interesting_ports_lookup is used, only the port is checked, not the protocol.
Plus, only the first result is returned.
This lookup include several entries for the same port (like 514 rsh/syslog) and the wrong entry can get returned.
This create false match. The wrong enrichment is given (including a wrong transport method) and a notable is generated when it shouldn't.

It can also be greatly optimized by moving the filter logic to the where of the tstats.

### **Expected behavior**
If a 514 udp is detected by the firewall, it should match syslog and not create a notable (or risk) telling me we detected a rsh 514 tcp.

### **App Version:**
 - ESCU: 5.19

### **Additional context**

Here is a solution (we get the transport from the log instead of adding it from the lookup, add transport to the match and move the filter logique up):
```
| tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime from datamodel=Network_Traffic where All_Traffic.action IN ("allowed", "allow") 
    [| inputlookup interesting_ports_lookup where is_prohibited="true" 
    | table dest_port transport 
    | dedup dest_port transport 
    | rename dest_port as All_Traffic.dest_port 
    | rename transport as All_Traffic.transport] by All_Traffic.src_ip All_Traffic.dest_ip All_Traffic.dest_port All_Traffic.action All_Traffic.dvc All_Traffic.src_port All_Traffic.vendor_product All_Traffic.rule All_Traffic.transport 
| lookup update=true interesting_ports_lookup dest_port as All_Traffic.dest_port transport as All_Traffic.transport OUTPUT app is_prohibited note 
| `drop_dm_object_name("All_Traffic")` 
| `prohibited_network_traffic_allowed_filter`
```

LaLaGuy:
### **Describe the bug**
The search can detect a 514 udp as rsh on tcp.
When "interesting_ports_lookup" is used, only the port is checked, not the protocol. Plus, only the first result is returned.
This lookup include several entries for the same port (like 514 rsh/syslog) and the wrong entry can get returned. This create false match. The wrong enrichment is given (including a wrong transport method) and a notable is generated when it shouldn't. It can also be greatly optimized by moving the filter logic to the where of the tstats.

### **Expected behavior**
If a 514 udp is detected by the firewall, it should match syslog and not create a notable (or risk) saying it detected a rsh 514 tcp.

### Details

It only contains a patch for the search in the rule "Prohibited Network Traffic Allowed".
It now get the transport from the log instead of adding it from the lookup, add transport to the match in the lookup "interesting_ports_lookup" command and the filter logic is now in the where.

### DIFF
--- a/detections/network/prohibited_network_traffic_allowed.yml
+++ b/detections/network/prohibited_network_traffic_allowed.yml
@@ -20,10 +20,10 @@ search: '| tstats `security_content_summariesonly` count min(_time) as firstTime
   where All_Traffic.action IN ("allowed", "allow") 
   [| inputlookup interesting_ports_lookup where is_prohibited="true" 
   | table dest_port transport | dedup dest_port transport 
-  | rename dest_port as All_Traffic.dest_port | rename transport as All_Traffic.transport]
+  | rename dest_port as All_Traffic.dest_port | rename transport as All_Traffic.transport] 
   by All_Traffic.src_ip All_Traffic.dest_ip All_Traffic.dest_port All_Traffic.action All_Traffic.dvc All_Traffic.src_port All_Traffic.vendor_product All_Traffic.rule 
-  | lookup update=true interesting_ports_lookup dest_port as All_Traffic.dest_port transport as All_Traffic.transport OUTPUT app is_prohibited note
-  | `security_content_ctime(firstTime)`
+  | lookup update=true interesting_ports_lookup dest_port as All_Traffic.dest_port transport as All_Traffic.transport OUTPUT app is_prohibited note 
+  | `security_content_ctime(firstTime)` 
   | `security_content_ctime(lastTime)` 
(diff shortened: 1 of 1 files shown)


ljstella:
Hi @LaLaGuy 

Thanks for the contribution! Just an FYI that the `appinspect` workflow is expected to fail on forks. 

We're currently going through a release process, but will give this the attention it deserves once we're done.

Thanks
Lou

ljstella:
@LaLaGuy would you mind increasing the version number and the date on the detection? We should be able to merge this once that's done.

LaLaGuy:
done

ljstella (approved):
LGTM. Thanks again for your contribution!

PR #3866, opened 2026-01-16
Update Analytics Performance 

### LINKED ISSUE #3809: Improving Performance of Detection Rules That Are Using Broad tstats Filters

nasbench:
This PR enhances the performance of a couple of detections as part of a continued effort that I was tracking from the issue #3809. Details are below.

> **Note All tests were executed on Endor**
> **First batch was addressed in https://github.com/splunk/security_content/pull/3850**

### Updated Analytics [11]

- `Common Ransomware Notes` - From 164 seconds to 33.44 seconds - Switch to using a subsearch since the lookup is small enough. Added a warning in the description to inform the user about this too.
- `File with Samsam Extension` - From 160 seconds to 11 seconds - Moved the extension search to the `where` and kept the extraction for better output.
- `Detect Rare Executables` - From 300~ seconds to 100~ seconds - Reduces usage of `values` and switched them to `latest` so that we can reduce MV calcs which are expensive.
- `3CX Supply Chain Attack Network Indicators` - Added a where clause with `DNS.query=*` to ensure the existence of the field, since it is used as a filter in the lookup. This is a best effort and will only reduce search time by some time. The same thing was applied to the following `Detect hosts connecting to dynamic domain providers`, `Detect Remote Access Software Usage DNS`
- `Detect Remote Access Software Usage URL` -  Added a where clause with `Web.url_domain=*` to ensure the existence of the field, since it is used as a filter in the lookup. 
- `Detect Remote Access Software Usage Traffic` - Added a where clause with `All_Traffic.app=*` to ensure the existence of the field, since it is used as a filter in the lookup.
- `Common Ransomware Extensions` - To enhance this we added filter for known extensions to reduce the search space. By excluding stuff like `.exe`...etc from file creation we already get rid of a lot. This can ofc be further enhanced by adding more to the list.
- `Detect Remote Access Software Usage File` - In order to also enhance perf, we allow only the extensions that are referenced by the lookup itself. Because this is using an file creation event, this will reduce the search space by a lot.
- `Windows DotNet Binary in Non Standard Path` - From 852 seconds to 7 seconds - Using a subsearch to enhance performance with the accepted limitation just because the lookup is small.

### DIFF
--- a/detections/network/detect_remote_access_software_usage_traffic.yml
+++ b/detections/network/detect_remote_access_software_usage_traffic.yml
@@ -1,7 +1,7 @@
 name: Detect Remote Access Software Usage Traffic
 id: 885ea672-07ee-475a-879e-60d28aa5dd42
-version: 11
-date: '2025-10-14'
+version: 12
+date: '2026-01-19'
 author: Steven Dick
 status: production
 type: Anomaly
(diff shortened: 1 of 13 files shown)


patel-bhavin (approved):

detections/network/detect_remote_access_software_usage_traffic.yml
```diff
@@ -15,16 +15,28 @@ description: The following analytic detects network traffic associated with
...
+  by All_Traffic.action All_Traffic.app All_Traffic.bytes All_Traffic.bytes_in
+     All_Traffic.bytes_out All_Traffic.dest All_Traffic.dest_ip
+     All_Traffic.dest_port All_Traffic.dvc All_Traffic.protocol
+     All_Traffic.protocol_version All_Traffic.src All_Traffic.src_ip
+     All_Traffic.src_port All_Traffic.transport All_Traffic.user
+     All_Traffic.vendor_product
+  | `drop_dm_object_name("All_Traffic")`
+  | `security_content_ctime(firstTime)`
+  | `security_content_ctime(lastTime)`
+  | lookup remote_access_software remote_appid AS app OUTPUT isutility, description as signature, comment_reference as desc, category
```
patel-bhavin, line 36: Should we move these lookup by the where clause like other detection updates? eg : Windows DotNet Binary in Non Standard Path

nasbench: We can't just because the app field is inconsistent in that CIM it can contain both apps categories and binaries, but more importantly the lookup is big with 600 entries, which will results in a big OR clause that results in the same performance.

patel-bhavin:
Lets wait on merging this after the release is out!

PR #3873, opened 2026-01-20
Fix Reported Issues

### LINKED ISSUE #3796: [BUG] ESCU - Suspicious Curl Network Connection issue
### **Describe the bug**
This detection is filtering based on the condition `Processes.process=s3.amazonaws.com`. This will never generate an alert if the datamodel is configured correctly as specified [here](https://help.splunk.com/en/data-management/common-information-model/6.0/data-models/endpoint). If you look at the description for Processes.process you'll see that it's intended to contain the name of the calling process so this would always fail to catch the intended acitivity. Changing this to `Processes.process IN("*s3.amazonaws.com*") ` would resolve this issue.

### **Expected behavior**
An alert/risk score is generated when a URL containing "s3.amazonaws.com" is curled.

### **App Version:**
- Splunk Cloud
- Enterprise Security Version: 8.2.2
- Build: 198157

### LINKED ISSUE #3805: Overlap Between “Executables Or Script Creation In Temp Path” and “Executables Or Script Creation In Suspicious Path”
Hi Splunk Team,

While reviewing the ESCU content, I noticed a functional overlap between the following two detections:

Executables Or Script Creation In Temp Path

Executables Or Script Creation In Suspicious Path

Both scenarios focus on detecting executable or script file creation in suspicious or abnormal locations. In practice, the Temp-path detection is a narrow subset of the more general suspicious-path rule. Because of this, they can be merged into a single consolidated detection without losing coverage.

Proposed Final Unified Scenario

Here is the combined search logic that accurately represents both use cases:

| tstats `security_content_summariesonly` values(Filesystem.file_path) as file_path count min(_time) as firstTime max(_time) as lastTime 
  from datamodel=Endpoint.Filesystem 
  where Filesystem.file_name IN ("*.exe", "*.dll", "*.sys", "*.com", "*.vbs", "*.vbe", "*.js", "*.ps1", "*.bat", "*.cmd", "*.pif", "*.msc")
    AND Filesystem.file_path IN ("*\\windows\\fonts\\*", "*\\windows\\temp\\*", "*\\users\\public\\*", "*\\windows\\debug\\*", "*\\Users\\Administrator\\Music\\*", "*\\Windows\\servicing\\*", "*\\Users\\Default\\*", "*Recycle.bin*", "*\\Windows\\Media\\*", "*\\Windows\\repair\\*", "*\\PerfLogs\\*", "*\\AppData\\Local\\Temp\\*")
  by Filesystem.action Filesystem.dest Filesystem.file_access_time Filesystem.file_create_time Filesystem.file_hash Filesystem.file_modify_time Filesystem.file_name Filesystem.file_path Filesystem.file_acl Filesystem.file_size Filesystem.process_guid Filesystem.process_id Filesystem.user Filesystem.vendor_product
| `drop_dm_object_name(Filesystem)`
| `executables_or_script_creation_in_suspicious_path_filter`

Additional Technical Note

The path C:\Windows\Temp is not a world-writable directory.
Because of this, treating it as a “world-writable high-risk location” in the filter logic of the original scenario is not accurate.
To avoid false assumptions and improve detection quality, this directory should not be categorized as world-writable.

Request

I recommend merging the two detections into one consolidated rule using the SPL above and adjusting the filter logic to remove the assumption that C:\Windows\Temp is world-writable.

Thank you! Let me know if any additional details are needed.

### LINKED ISSUE #3812: [BUG] Noisy Filter in Windows Registry Delete Task SD
in this scenario the default filter is :
| tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime FROM datamodel=Endpoint.Registry where Registry.registry_path IN ("*\\Schedule\\TaskCache\\Tree\\*") Registry.user="SYSTEM" Registry.registry_value_name="SD" (Registry.action=Deleted OR Registry.action=modified) by Registry.action Registry.dest Registry.process_guid Registry.process_id Registry.registry_hive Registry.registry_path Registry.registry_key_name Registry.registry_value_data Registry.registry_value_name Registry.registry_value_type Registry.status Registry.user Registry.vendor_product 
| `drop_dm_object_name(Registry)` 
| `windows_registry_delete_task_sd_filter`

the Registry.action filter "modify" is very noisy and the rule will fire with each scheduled task creation when we create a new scheduled task the value of this three reg paths will modify as follow 
HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Schedule\TaskCache\Tree\TEST1\Index
HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Schedule\TaskCache\Tree\TEST1\Id
HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Schedule\TaskCache\Tree\TEST1\SD
so please remove the Registry.action=modified from the rule
the picture below show the sysmon event log after creating a scheduled task called test1

Image

### LINKED ISSUE #3828: [BUG] Logic Problem of Process Creating LNK file in Suspicious Location scenario
Hi in my opinion the logic of Process Creating LNK file in Suspicious Location scenario have problem and beacuse of this problems it cause huge amount of FPs 
the first problem is the lak of Filesystem.image filed to determine which binary create this lnk file
after that in every file handle based that achived by a process a lnk file will created based on this handle achivement
so lets dissect the created file structure based on my opinion 
image we open a excel file called mahdi.csv based on the handle that excel.exe process achived on mahdi.csv this file will be create --> mahdi.csv.lnk
so the first thing we have to do is extract the real extension based on it 
the real extension is csv so first we have to achive the real extension with regex after that exclude the expected binray and extension like 
| search NOT image="*\\exel.exe" AND extension=".csv"
without this logic the investigation of this scenario is useless.

### LINKED ISSUE #3869: [BUG]ESCU – PaperCut NG Remote Web Access Attempt: High Noise Due to Generic /app Endpoint Usage
Issue Summary

The current detection logic treats the /app endpoint as an independent indicator of suspicious activity. However, /app is a generic endpoint used by many different web applications, not exclusively by PaperCut NG.

Filtering or alerting on this endpoint in isolation leads to excessive false positives, as legitimate traffic from unrelated applications frequently matches this pattern.

Impact

High false-positive rate

Significant alert noise in environments hosting multiple web applications

Reduced analyst confidence in the detection

Increased triage time for non-malicious events

Technical Rationale

/app is a commonly used endpoint across a wide range of web frameworks and products.

Many non-PaperCut applications legitimately expose or access this path.

The rule does not sufficiently validate that the traffic is uniquely associated with PaperCut NG before triggering.

Without additional context (such as application-specific paths, exploit indicators, or version-aware logic), /app alone is not a reliable signal of exploitation.

Recommendation

Remove /app as a standalone detection condition.

Scope the rule more narrowly to PaperCut NG–specific indicators.

Correlate web access with stronger signals, such as:

Known PaperCut exploit paths

Abnormal request parameters

Authentication bypass behavior

Confirmed vulnerable versions

Suspicious HTTP methods or response patterns

Expected Outcome

Lower false-positive rates

Improved detection fidelity

More actionable alerts for real PaperCut NG exploitation attempts.

nasbench:
This PRs aims to fix a couple of the reported issues. Details are below

### Updated Analytics [36]

- `Executables Or Script Creation In Suspicious Path` - Removed `temp` path from this detection as it does not fit. Instead moved it to `Executables Or Script Creation In Temp Path`, I have also increased the risk score to `30` and enhanced a bit of the metadata.
- `Executables Or Script Creation In Temp Path` - Added a generic filter for `__PSScriptPolicyTest_` to reduce FP. And reduced the score to `5`. Since this will produce a lot of FPs, because of the `.exe`, `.dll` extensions from some installers. 
- `Linux Ingress Tool Transfer with Curl` - Added an optimization where we add a filter condition for the flag in the tstats to reduce the search scope later on with the regex.
- `Process Creating LNK file in Suspicious Location` - Changed this to an `Anomaly` and added multiple default filters to reduce noise. Since the rule is looking for any shortcut creation in the users folders, this could be super noisy. Especially for files located in `\Users\Public`, `\Desktop`, `\Downloads`, etc. I have also reduced the risk score to reflect this and added missing TO.
- `Suspicious Curl Network Connection` - Fixed a bug where the commandline was missing wildcards.
- `Windows Registry Delete Task SD` - Fixed a bug where the logic was looking for modifications, which is outside the scope of the detection (according to the metadata/title). Also the previous `action` and `value` filters were incorrect.
- Fix the data for all the `Cisco Secure Firewall` detections.
- `PaperCut NG Remote Web Access Attempt` - Removed the `/app` endpoint from the logic due to high FP report from community.

### Breaking Change [1]

- `Process Creating LNK file in Suspicious Location` - Changing this to an `Anomaly` means it will not produce a finding anymore but only a IF.


detections/web/papercut_ng_remote_web_access_attempt.yml
```diff
@@ -1,75 +1,112 @@
...
+  from datamodel=Web WHERE
+
+  Web.url IN (
+    "/app?service=direct/1/PrinterDetails/printerOptionsTab.tab"
+    "/app?service=direct/1/PrinterList/selectPrinter&sp=*",
+    "/app?service=page/PrinterList",
+    "/app?service=page/SetupCompleted"
+  )
+  NOT src IN (
+                "10.0.0.0/8",
```
patel-bhavin, line 30: curious if it makes sense to add `internal_subnets` macro? i have noticed this added to multiple searches

nasbench: We can make that optimization. Let me do it

nasbench: After thinking, let me do this once this and the other PRs are merged. Since there are references to this in both (and many other places.)

patel-bhavin: that makes sense!

patel-bhavin (approved):

PR #3874, opened 2026-01-21
Additional information in Risk Message for Services LOLBAS spawn detection

DipsyTipsy:
### Details

Adding information of what LOLBAS has been spawned to the risk message as this gives a quick reference into the activity observed.

### Checklist

- [ ] Validated SPL logic.

### DIFF
--- a/detections/endpoint/services_lolbas_execution_process_spawn.yml
+++ b/detections/endpoint/services_lolbas_execution_process_spawn.yml
@@ -67,7 +67,7 @@ drilldown_searches:
   earliest_offset: $info_min_time$
   latest_offset: $info_max_time$
 rba:
-  message: Services.exe spawned LOLBAS: $process_name$ on $dest$
+  message: "Services.exe spawned LOLBAS: $process_name$ on $dest$"
   risk_objects:
   - field: dest
     type: system
--- a/detections/endpoint/services_lolbas_execution_process_spawn.yml
+++ b/detections/endpoint/services_lolbas_execution_process_spawn.yml
@@ -67,7 +67,7 @@ drilldown_searches:
   earliest_offset: $info_min_time$
   latest_offset: $info_max_time$
 rba:
-  message: Services.exe spawned a LOLBAS process on $dest$
+  message: Services.exe spawned LOLBAS: $process_name$ on $dest$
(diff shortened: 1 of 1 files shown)


detections/endpoint/services_lolbas_execution_process_spawn.yml
```diff
@@ -67,7 +67,7 @@ drilldown_searches:
   earliest_offset: $info_min_time$
   latest_offset: $info_max_time$
 rba:
-  message: Services.exe spawned a LOLBAS process on $dest$
+  message: "Services.exe spawned LOLBAS: $process_name$ on $dest$"
```
nasbench: Small suggestion. Also if you could add those along with `process` as threat objects. That would be great. Since you already making the changes.

Thanks

```suggestion
  message: "Services.exe spawned LOLBAS: $process_name$ located in $process_path$ on $dest$"
```

detections/endpoint/services_lolbas_execution_process_spawn.yml
nasbench: We also need to up the date and version

detections/endpoint/services_lolbas_execution_process_spawn.yml
DipsyTipsy: Good catch, I'll update my PR! Should we use `process` or `process_name`?  Seems like `process_name` is more used across the repo.

nasbench: I like both, process is useful to see in threat topology as it gives you a good pivot idea. But i have no strong preference since as you said we mainly use porcess_name.

DipsyTipsy: An other question, in my checks I came across several other lolbas queries. Most of these should probably get the same change. Should I just add that to this PR?

Bonus is that they all have the various lolbins declared in the spl, we could either move that to a macro or utilize https://github.com/splunk/security_content/blob/develop/lookups/lolbas_file_path.csv, to reduce the maintenance a bit.

nasbench (approved):
Changes LTGTM. Thanks Emil

patel-bhavin (approved):

PR #3876, opened 2026-01-22
Updating query based

patel-bhavin:
updated - DNS Query Length With High Standard Deviation

Based on testing various combinations of threshold values, the current SPL gives the best results! 
Also removed"SOA", "SRV") DNS.query != *. queries which also resulted in reducing result count from 1.9k to 41 events


nasbench (approved):
LGTM

PR #3880, opened 2026-01-29
Update RBA, logic, and beautify some searches

nasbench:
This PR updates the logic and RBA of a couple of searches

### Updated Analytics [3]

- `Internal Horizontal Port Scan NMAP Top 20` - Add new local IP ranges.
- `Internal Vertical Port Scan` - Add new local IP ranges and revert back RBA objects config.
- `Protocol or Port Mismatch` - Add new ports for mismatch related to SMTP and its TLS use.


detections/network/internal_vertical_port_scan.yml
```diff
@@ -49,11 +79,11 @@ drilldown_searches:
 rba:
   message: $src_ip$ has scanned $totalDestPortCount$ ports on $dest_ip$
   risk_objects:
-  - field: dest_ip
+  - field: src_ip
     type: system
     score: 60
   threat_objects:
-  - field: src_ip
+  - field: dest_ip
```
patel-bhavin, line 86: lets perhaps remove "dest_ip" from threat_objects, much like the above detection

nasbench: ES expect a TO for investigation, and `dest_ip` make sense here. Don't you think we should instead add it to the other rule?

nasbench: As discussed let's keep it for now

patel-bhavin (approved):

PR #3911, opened 2026-02-18
AITK?

patel-bhavin:
Since MLTK content was deprecated in https://github.com/splunk/security_content/pull/3922 and will be removed in v.5.26.0. We not longer host MLTK/AITK content in this repo. Hence this PR is removing the presence of their respective TA from the contentctl.yml


ljstella:
Marking WIP so we don't accidentally merge even if this "passes", as these dependencies need to be special-cased.

ljstella:
Potentially blocking here on the splunk/splunk container version, and the included python - but internal tests so far have showed this works as-is with PSC 4.2.5

nasbench (approved):

nasbench:
Switching back to draft as to not accidentally merge. See internal discussion, this PR is ready to merge by v5.26

nasbench (approved):

PR #3925, opened 2026-02-26
ta update script

P4T12ICK:
This PR contains a CI workflow which checks for TA used by our detections, download it and put it into the attack-range-appbinaries S3 bucket


patel-bhavin:
checks out good locally , merging it so that we can check its cron run at midnight!

patel-bhavin (approved):

patel-bhavin (approved):

PR #3929, opened 2026-03-02
Split and tweak of  linux_docker_privilege_escalation

DipsyTipsy:
### Details

Current logic in linux_docker_privilege_escalation hits on a lot of undesired activity because of the generous wildcards. The query in it self also tries to perform a bit to many things in one single query. Suggesting splitting the detection into two:

- linux_docker_root_directory_mount (Suggesting moving to TTP)
- linux_docker_shell_execution 

Tweaked the queries to hit more precisely, as well as updating the mitre annotations.
Not entirely sure how you want the retirement of detection's to be performed, so leaving this out.  Also a bit unsure on how the versioning and author values should be handled in these cases.

### Checklist

- [ ] Validated SPL logic.


nasbench (commented):
Just general comments before review based on your question. 

- The new detections will have new UUIDs/Version.
- The old detection should be moved to the deprecated folder in the detections folder.
- We need to update the deprecation mapping yaml
- As for the author, if the search is was not changed just a split. Use only the original author. If you did some changes to the logic then add yourself.

I will give this a review later on. Thanks for taking the time on this,

Do not worry on the deprecation stuff. We will do it on your behalf.

nasbench (approved):
LGTM. I deprecated the other detections and tweaked formatting as well as the list of shells.

Since you can execute the shell as the last param to start a new shell session.

patel-bhavin:
Looks good @nasbench, @DipsyTipsy ! :shipit: Thanks for shipping a clear breakdown for these detections!

PR #3930, opened 2026-03-02
Remove EXTRA fields

pyth0n1c:
A number of ymls have extra fields that are not part of the model. 
This should be a contentctl validate exception, but is not.

I have intentionally not bumped the versions/dates for these detections because these changes should not impact how the content is serialized into CONF file, meaning they should not require a new version in the metadata field for ES8 versioning purposes.


ljstella (approved):
LGTM

PR #3931, opened 2026-03-02
normalize all legacy line endings

pyth0n1c:
Fix line endings that used windows format CRLF to normalized LF


pyth0n1c:
I am reasonably sure that the failures noted in the unit-testing workflow at this time are due to lingering issues with the `develop` branch that this PR will fix.
Also note that the file diff in GitHub is not particularly useful - it does not show the difference between CRLF and LF in the Web UI

nasbench (approved):

nasbench:
Merging this as its just EOL changes. We'll check the CI afterwards.

PR #3946, opened 2026-03-10
anomaly_standard_init_score

tccontre:
This PR is to implement the STRT initial RBA score for detection type Anomaly : 20

Screenshot 2026-03-10 at 09 40 22

total detection type updated: {'Anomaly': 730}


patel-bhavin (approved):

patel-bhavin:
🚢  - Lets also include them as a check in contentctl:

Scores allowed : 50 for TTP, 20 for Anomaly!

cc: @pyth0n1c @cmcginley-splunk

pyth0n1c:
> 🚢 - Lets also include them as a check in contentctl:
> 
> Scores allowed : 50 for TTP, 20 for Anomaly!
> 
> cc: @pyth0n1c @cmcginley-splunk

To clarify, this would mean that:
(fixed for clarity after Lou's comment)


**- TTP always 50**
**- Anomaly always 20**



So this would NOT be a default, but instead something that cannot be overridden.
Is that correct?

patel-bhavin:
> > 🚢 - Lets also include them as a check in contentctl:
> > Scores allowed : 50 for TTP, 20 for Anomaly!
> > cc: @pyth0n1c @cmcginley-splunk
> 
> To clarify, this would mean that anomaly MUST ALWAYS have 50 and TTP MUST ALWAYS have 20 right? So this would NOT be a default, but instead something that cannot be overridden. Is that correct?

Yes, that is correct

ljstella:
@patel-bhavin @pyth0n1c other way around on the scores, I believe- TTP gets 50, anomaly gets 20

pyth0n1c:
> @patel-bhavin @pyth0n1c other way around on the scores, I believe- TTP gets 50, anomaly gets 20

Thanks, my mixup. I updated my original comment to reflect this!

Issue #3954, opened 2026-03-13
[BUG] src_user field in Azure AD PIM Role Assigned


AndreiBanaru:
The [Azure AD PIM Role Assigned](https://github.com/splunk/security_content/blob/v5.23.0/detections/cloud/azure_ad_pim_role_assigned.yml) detection calls `fillnull `and then aggregates over some fields, including `src`.

We are on version 6.1.0 of the `Splunk Add-on for Microsoft Cloud Services`, with this present in the `props.conf` file (I've stripped unimportant stuff):

```editorconfig
[azure:monitor:aad]
FIELDALIAS-src = callerIpAddress as src
EVAL-src_user = case(operationName IN ("Provisioning activity"),'properties.initiatedBy.Name', \
                        operationName IN ("Add member to role","Update user","Add eligible member to role","Remove eligible member from role"),'properties.initiatedBy.app.displayName', \
                        operationName IN ("Delete user","Remove member from role"),if('properties.initiatedBy.app.servicePrincipalId'!="null",'properties.initiatedBy.app.servicePrincipalId','properties.initiatedBy.app.displayName'), \
                        'properties.initiatedBy.user.userPrincipalName'!="",'properties.initiatedBy.user.userPrincipalName', \
                        'properties.initiatedBy.user.displayName'!="",'properties.initiatedBy.user.displayName')
```

In the raw events, there's no `callerIpAddress` to create the `src` field from, so this will always be `0`, since you call `fillnull`.

In our environment I replaced `src` with `src_user` as I believe this aligns more to the detections scope (e.g. who assigned the role).

patel-bhavin:
that checks out! lets replace src with src_user for this deection.

PR #3955, opened 2026-03-13
Replace src with src_user in Azure AD PIM Role Assigned

AndreiBanaru:
As described in #3954, this PR will replace `src` with `src_user`.


patel-bhavin (approved):

patel-bhavin:
thanks @AndreiBanaru for creating this fix! app inspect expected to fail on PRs from forks but the change LGTM!

PR #3962, opened 2026-03-18
Fix Reported Issues - April Batch

### LINKED ISSUE #3909: [BUG]Detection Overlap Between “Windows Scheduled Task Service Spawned Shell” and “Possible Lateral Movement Powershell Spawn
While testing detection content in Enterprise Security Content Update (ESCU) from Splunk, I observed significant overlap between the following two analytics:

ESCU – Windows Scheduled Task Service Spawned Shell

ESCU – Possible Lateral Movement Powershell Spawn

Both detections trigger when powershell.exe (or another shell) is spawned by svchost.exe.

Because the Windows Task Scheduler service runs under svchost.exe, legitimate scheduled task executions frequently satisfy the parent-process condition in both rules. This results in duplicate alerts for the same event.

Problem

svchost.exe is a generic Windows service host process that runs numerous services, including:

Task Scheduler (Schedule)

WinRM

RPC

Various system services

Using svchost.exe as a broad parent-process indicator in the Possible Lateral Movement Powershell Spawn analytic reduces precision and causes:

Duplicate detections

Analyst fatigue

Reduced clarity between persistence vs lateral movement scenarios

Increased false positive volume in environments with legitimate scheduled tasks

The current logic makes it difficult to clearly separate:

Scheduled task–based execution

Service-based execution

Actual remote/lateral PowerShell activity

Recommended Action

I recommend one of the following:

Option 1 – Deprecate the “Windows Scheduled Task Service Spawned Shell” analytic

If the lateral movement analytic is intended to cover this behavior more generically, the scheduled task–specific rule may be redundant and could be deprecated.

Option 2 – Remove svchost.exe from the parent condition in “Possible Lateral Movement Powershell Spawn”

This would:

Eliminate the overlap

Preserve analytic separation of concerns

Improve detection fidelity

Reduce duplicate alerts

If retaining svchost.exe is necessary, additional context (service group filtering, remote execution indicators, or network telemetry correlation) should be required instead of using it as a standalone suspicious parent.

### LINKED ISSUE #3939: [BUG] Malicious PowerShell Process - Encoded Command - regex doesn't make sense
### **Describe the bug**
The regex used to matching command line argument makes no sense

### **Expected behavior**
The regex should match a string containing a "-encodedcommand something" (with encodedcommand parameter given with various dashes or slash) sequence but not match other ones.

### **App Version:**
- Irrelevant

### **Additional context**
There are several issues with the regex string.

`\"(?i)[\\-|\\/|–|—|―][Ee^]{1,2}[NnCcOoDdEeMmAa^]+\\s+[\\\"]?[A-Za-z0-9+/=]{5,}[\\\"]?\"`
Firstly we need to unescape it since it's in a yaml so in order to get the original regex we need to remove some backslashes.
We end up with 
`"(?i)[\-|\/|–|—|―][Ee^]{1,2}[NnCcOoDdEeMmAa^]+\s+[\"]?[A-Za-z0-9+/=]{5,}[\"]?"`

1. Since we declare the regex case-insensitive by using (?i) there is no point in listing all letters in both upper and lowercase
2. The characters in character lists don't have to be escaped.
3. In the character list you don't need pipes to specify alternatives so the first character list is way too long. Also it makes the rule match "|encodedcommand" as well
4. It won't capture the case when someone passes the encoded command as a variable.
5. There's not much point in making the quotes single-character lists

So the regex can be tweaked to
`"(?i)[-/–—―][e^]{1,2}[ncodema^]+\s+\"?[a-z0-9+=]{5,}\"?"`
And extended to catch the variable use
`"(?i)[-/–—―][e^]{1,2}[ncodema^]+\s+\"?([a-z0-9+=]{5,}|\$.+)\"?"`
(unfortunately, variable names can be made of almost any characters)

### LINKED ISSUE #3961: [BUG] Logic Problem in Detect Computer Changed with Anonymous Account
the detection rule original logic is as follow: 
`wineventlog_security` EventCode=4624 OR EventCode=4742 TargetUserName="ANONYMOUS LOGON" LogonType=3
    BY action app authentication_method
       dest dvc process
       process_id process_name process_path
       signature signature_id src
       src_port status subject
       user user_group vendor_product
| `detect_computer_changed_with_anonymous_account_filter`

there are two problem with this logic.
first problem is filtering EventCode=4742 does not effect in the query beacuse of the LogonType=3 field this field is not available in Event Code 4742 telemetry, so the Event Code 4742 doesnt work.

second problem is that the event 4624 filter is extra and not needed.

### LINKED ISSUE #3967: [BUG] Data sources reference obsolete/archived Palo Alto add-on
### **Describe the bug**
The `data_sources/palo_alto_network_traffic.yml` and `data_sources/palo_alto_network_threat.yml` reference the old, archived "Palo Alto Networks Add-on" (https://splunkbase.splunk.com/app/2757). 

### **Expected behavior**
Update the data sources to reference the new Palo Alto add-ons, "Splunk Add-on for Palo Alto Networks" and/or " Splunk App for Palo Alto Networks"

### **Screenshots**
Image

### **App Version:**
 - ESCU: 5.23.0

### LINKED ISSUE #3969: Add Event ID 4756 to windows_privileged_group_modification detection
**Is your feature request related to a problem? Please describe.**
Yes.
While using the windows_privileged_group_modification detection in production, we observed that additions to highly privileged Active Directory groups (notably Enterprise Admins) are logged using Event ID 4756 (A member was added to a security-enabled universal group).
The current detection only includes Event IDs 4727, 4731, 4744, 4749, 4754, 4759, 4783, 4790, which results in missed detections for real-world privilege escalation scenarios involving universal security groups such as Enterprise Admins.

**Describe the solution you'd like**
Add Event ID 4756 to the SPL used in the windows_privileged_group_modification detection.

**Describe alternatives you've considered**
As a workaround, we locally modified the detection to include Event ID 4756.

**Additional context**
Event ID 4756 is the standard Windows Security Event generated when a member is added to a security-enabled universal group and is logged on Domain Controllers. This behavior is consistently observed in production Active Directory environments. Missing this Event ID can allow high-impact privilege escalations (e.g. addition to Enterprise Admins) to go undetected.

nasbench:
This PR update and fixes multiple discovered and reported issues. Below is a summary of the changes.

### New Data Sources

- `Windows Event Log Security 4756` - Added this new DS to link to a related detection.

### Updated Data Sources

- `Palo Alto Network Threat` - Updated this DS to use the `Palo Alto Networks Add-on`
- `Palo Alto Network Traffic` - Updated this DS to use the `Palo Alto Networks Add-on`

### Updated Analytics

- `Email Attachments With Lots Of Spaces` - Added missing MITRE IDs.
- `Suspicious Java Classes` - Fix broken regex (from last YAML script application) and added missing MITRE IDs.
- `Cloud Compute Instance Created With Previously Unseen Instance Type` - Added missing MITRE IDs.
- `Certutil exe certificate extraction` - Added missing MITRE IDs.
- `Malicious PowerShell Process - Encoded Command` - Updated the broken regex with a more robust one that aims to detect most variation of the `EncodedCommand` flag (Fix #3939)
- `Outbound Network Connection from Java Using Default Ports` - Remove duplicate entry for `javaw.exe` and other updates to the SPL structure so that its more readable.
- `Suspicious Rundll32 no Command Line Arguments` - Remove the unnecessary usage of regex and moved the filter logic earlier for better performance
- `Suspicious SearchProtocolHost no Command Line Arguments` - Remove the unnecessary usage of regex and moved the filter logic earlier for better performance
- `Windows New Deny Permission Set On Service SD Via Sc.EXE` - Updated metadata info, including the FP section based on Athena telemetry.
- `Windows New Service Security Descriptor Set Via Sc.EXE` - Updated metadata info, including the FP section based on Athena telemetry.
- `Windows Privilege Escalation Suspicious Process Elevation` / `Windows Privilege Escalation User Process Spawn System Process` / `Windows UAC Bypass Suspicious Escalation Behavior` - Beautified the SPL of these 3 analytics for better reading  (considerations are being made to deprecated these due to their poor performance).
- `Detect Large ICMP Traffic` / `Detect Outbound LDAP Traffic` - Update the logic to these by adding a more broad filter for local IPs.
- Beautified the SPL of multiple analytics that were leveraging the Palo Alto TA.
- `Detect Computer Changed with Anonymous Account` - Updated the logic to be more accurate. (See explanation in #3961)
- `Windows Privileged Group Modification` - Update logic to include EventID 4756 (Fix #3969)
- `Windows Scheduled Task Service Spawned Shell` - Update and beautify the SPL as well as other metadata and RBA related config.
- `Possible Lateral Movement PowerShell Spawn` - Fixed FP by adding exclusion for svchost with the schedule service


data_sources/windows_event_log_security_4756.yml
```diff
@@ -0,0 +1,18 @@
...
+separator: EventCode
+supported_TA:
+- name: Splunk Add-on for Microsoft Windows
+  url: https://splunkbase.splunk.com/app/742
+  version: 9.1.2
+fields:
+- _time
+output_fields:
+- dest
+example_log: ''
```
patel-bhavin: lets add an example log here?

nasbench: Added one. I copied this from the other templates. Hence its minimal with no real effort behind it.

patel-bhavin:
neat set of changes and PR description! LGTM 💯

patel-bhavin (approved):

PR #3972, opened 2026-03-26
Update detections from Athena FPs

patel-bhavin:
added filters to detections to reduce FPs from athena runs!

### DIFF
--- a/detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
+++ b/detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
@@ -4,7 +4,7 @@ version: 14
 date: '2026-03-24'
 author: Bhavin Patel, Mauricio Velazco, Splunk
 status: production
-type: TTP
+type: Anomaly
 description: The following detects the execution of cscript.exe or wscript.exe processes spawned by cmd.exe, leveraging Endpoint Detection and Response (EDR) telemetry mapped to the Endpoint data model, with additional contextual filtering to improve fidelity and reduce false positives. It focuses on executions originating from user-writable directories such as Users, AppData, Temp, and Downloads, which are commonly abused by attackers to stage and execute malicious scripts, while excluding trusted system paths like C:\Windows\System32\ and C:\Program Files\ that are typically associated with legitimate activity. The detection also filters out service accounts (e.g., accounts ending with $ or known naming conventions) to minimize noise from automated processes and incorporates command-line context to better assess script execution patterns and identify potentially suspicious behavior. These enhancements transform the analytic from a broad pattern-based detection into a more behavior-driven approach, increasing its effectiveness in identifying malicious script interpreter usage that could indicate initial access, persistence, or lateral movement.
 data_source:
     - Sysmon EventID 1
--- a/detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
+++ b/detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
@@ -5,7 +5,10 @@ date: '2026-03-24'
 author: Bhavin Patel, Mauricio Velazco, Splunk
 status: production
 type: Anomaly
-description: The following detects the execution of cscript.exe or wscript.exe processes spawned by cmd.exe, leveraging Endpoint Detection and Response (EDR) telemetry mapped to the Endpoint data model, with additional contextual filtering to improve fidelity and reduce false positives. It focuses on executions originating from user-writable directories such as Users, AppData, Temp, and Downloads, which are commonly abused by attackers to stage and execute malicious scripts, while excluding trusted system paths like C:\Windows\System32\ and C:\Program Files\ that are typically associated with legitimate activity. The detection also filters out service accounts (e.g., accounts ending with $ or known naming conventions) to minimize noise from automated processes and incorporates command-line context to better assess script execution patterns and identify potentially suspicious behavior. These enhancements transform the analytic from a broad pattern-based detection into a more behavior-driven approach, increasing its effectiveness in identifying malicious script interpreter usage that could indicate initial access, persistence, or lateral movement.
+description: |
+    The following detects the execution of cscript.exe or wscript.exe processes spawned by cmd.exe, leveraging Endpoint Detection and Response (EDR) telemetry mapped to the Endpoint data model, with additional contextual filtering to improve fidelity and reduce false positives.
+    It focuses on executions originating from user-writable directories such as Users, AppData, Temp, and Downloads, which are commonly abused by attackers to stage and execute malicious scripts, while excluding trusted system paths like C:\Windows\System32\ and C:\Program Files\ that are typically associated with legitimate activity.
+    The detection also filters out service accounts (e.g., accounts ending with $ or known naming conventions) to minimize noise from automated processes and incorporates command-line context to better assess script execution patterns and identify potentially suspicious behavior.
 data_source:
     - Sysmon EventID 1
     - Windows Event Log Security 4688
--- a/detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
+++ b/detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
@@ -14,34 +14,44 @@ data_source:
     - Windows Event Log Security 4688
     - CrowdStrike ProcessRollup2
 search: |-
-    | tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime
-    FROM datamodel=Endpoint.Processes
-    WHERE Processes.parent_process_name="cmd.exe"
-      AND (Processes.process_name="cscript.exe" OR Processes.process_name="wscript.exe")
-      AND NOT Processes.process IN (
-        "* C:\\Windows\\System32\\*",
-        "* \"C:\\Windows\\System32\\*",
-        "* C:\\Windows\\SysWOW64\\*",
-        "* \"C:\\Windows\\SysWOW64\\*",
-        "* C:\\Program Files\\*",
+    | tstats `security_content_summariesonly`
+      count min(_time) as firstTime
+            max(_time) as lastTime
+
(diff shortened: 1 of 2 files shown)


detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
```diff
@@ -1,33 +1,36 @@
 name: Detect Use of cmd exe to Launch Script Interpreters
 id: b89919ed-fe5f-492c-b139-95dbb162039e
-version: 13
-date: '2026-03-10'
+version: 14
+date: '2026-03-24'
 author: Bhavin Patel, Mauricio Velazco, Splunk
 status: production
 type: TTP
```
nasbench: This needs to be reduced to Anomaly. This is a super common activity in general. 

```suggestion
type: Anomaly
```

detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
```diff
@@ -1,33 +1,36 @@
...
+      Processes.parent_process Processes.parent_process_exec Processes.parent_process_guid
+      Processes.parent_process_id Processes.parent_process_name Processes.parent_process_path
+      Processes.process Processes.process_exec Processes.process_guid
+      Processes.process_hash Processes.process_id Processes.process_integrity_level
+      Processes.process_name Processes.process_path Processes.user
+      Processes.user_id Processes.vendor_product
     | `drop_dm_object_name("Processes")`
+    | where NOT like(process, "%\\Windows\\System32\\%")
+    | where NOT like(process, "%\\Program Files\\%")
+    | where NOT match(user, ".*\\$$")
```
nasbench: Move these filters earlier for better performance.

Also add other entries like `\Windows\SySWOW64` and `Program Files (x86)` and anchor them with `C:` to avoid blind spots related to similar folder names.

Also filtering like this on CLI is dangerous since, cscript.exe could be invoked like 

`C:\Windows\System32\cscript.exe .....some...commandline....`. And by simply filtering on the string `\Windows\System32\` You exclude all entries.

I suggest a filter like this 

```
NOT Processes.process IN (
    "* C:\\Windows\\System32\\",
    "* \"C:\\Windows\\System32\\",
    "* C:\\Windows\\SysWOW64\\",
    "* \"C:\\Windows\\SysWOW64\\",
    "* C:\\Program Files\\",
    "* \"C:\\Program Files\\",
    "* C:\\Program Files (x86)\\",
    "* \"C:\\Program Files (x86)\\",
)
```

The extra space avoid matches with the actual binary, and the quote version include matches that use quoted paths.

detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
```diff
@@ -1,33 +1,36 @@
...
+description: The following detects the execution of cscript.exe or wscript.exe processes spawned by cmd.exe, leveraging Endpoint Detection and Response (EDR) telemetry mapped to the Endpoint data model, with additional contextual filtering to improve fidelity and reduce false positives. It focuses on executions originating from user-writable directories such as Users, AppData, Temp, and Downloads, which are commonly abused by attackers to stage and execute malicious scripts, while excluding trusted system paths like C:\Windows\System32\ and C:\Program Files\ that are typically associated with legitimate activity. The detection also filters out service accounts (e.g., accounts ending with $ or known naming conventions) to minimize noise from automated processes and incorporates command-line context to better assess script execution patterns and identify potentially suspicious behavior. These enhancements transform the analytic from a broad pattern-based detection into a more behavior-driven approach, increasing its effectiveness in identifying malicious script interpreter usage that could indicate initial access, persistence, or lateral movement.
```
nasbench: This is not needed `These enhancements transform the analytic from a broad pattern-based detection into a more behavior-driven approach, increasing its effectiveness in identifying malicious script interpreter usage that could indicate initial access, persistence, or lateral movement.`

So here is a suggestion to remove it.


```suggestion
description: |
    The following detects the execution of cscript.exe or wscript.exe processes spawned by cmd.exe, leveraging Endpoint Detection and Response (EDR) telemetry mapped to the Endpoint data model, with additional contextual filtering to improve fidelity and reduce false positives.
    It focuses on executions originating from user-writable directories such as Users, AppData, Temp, and Downloads, which are commonly abused by attackers to stage and execute malicious scripts, while excluding trusted system paths like C:\Windows\System32\ and C:\Program Files\ that are typically associated with legitimate activity.
    The detection also filters out service accounts (e.g., accounts ending with $ or known naming conventions) to minimize noise from automated processes and incorporates command-line context to better assess script execution patterns and identify potentially suspicious behavior.
```

detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
```diff
@@ -41,12 +44,14 @@ drilldown_searches:
       earliest_offset: $info_min_time$
       latest_offset: $info_max_time$
 rba:
-    message: cmd.exe launching script interpreters $process_name$ on $dest$
+    message: cmd.exe launching script interpreters [$process_name$] with commandline [$process$] on [$dest$]
```
nasbench:
```suggestion
    message: cmd.exe launched a script interpreter [$process_name$] with CommandLine [$process$] on [$dest$]
```

detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
```diff
@@ -1,16 +1,37 @@
 name: Scheduled Task Deleted Or Created via CMD
 id: d5af132c-7c17-439c-9d31-13d55340f36c
-version: 25
-date: '2026-03-10'
+version: 26
+date: '2026-03-26'
 author: Bhavin Patel, Splunk
 status: production
 type: TTP
```
nasbench: This also needs to be brought down to an Anomaly
```suggestion
type: Anomaly
```

detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
```diff
@@ -1,16 +1,37 @@
...
+        min(_time) as firstTime
+        max(_time) as lastTime
+    FROM datamodel=Endpoint.Processes
+    WHERE
+        Processes.process_name="schtasks.exe"
+        AND (
+            Processes.process IN ("*/create*","*-create*","*/delete*","*-delete*" )
+        )
+        AND (
+            Processes.process IN ("*/tr*","*/sc*","*/ru*")
```
nasbench: Why are you adding these flags? If the aim is to flag any create/delete then they are not needed.

Also, in their current state the `delete` flag will never be matched. Since these flags are not available when its used.

detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
```diff
@@ -1,16 +1,37 @@
...
+        )
+        AND NOT Processes.process IN ("*\\Windows\\System32\\*","*\\Program Files\\*")
```
nasbench: Same logic with my above comment around the filter. Please update this as well.

detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
```diff
@@ -1,16 +1,37 @@
...
+        AND Processes.parent_process_name IN ("cmd.exe","wscript.exe","cscript.exe")
```
nasbench: The analytics is stating via cmd in the title. So either update the title or remove the scripting entries.

detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
nasbench, line 59: Add links to schtasks.


```suggestion
- https://thedfirreport.com/2022/02/21/qbot-and-zerologon-lead-to-full-domain-compromise/
- https://www.joesandbox.com/analysis/691823/0/html
- https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/schtasks-delete
- https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/schtasks-create
```

nasbench: Add `process` as a threat object

nasbench: Update the RBA message to include the parent launching this with its command line

nasbench (changes requested):

detections/endpoint/detect_use_of_cmd_exe_to_launch_script_interpreters.yml
patel-bhavin: @nasbench - thank you Nas! Moved the filters earlier and added full paths! Appreciate the commit suggestions too!

detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
patel-bhavin: done

detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
patel-bhavin: done!

detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
patel-bhavin: neat point that some of the delete commands would be ignored! i removed that additional constraint in process args

detections/endpoint/scheduled_task_deleted_or_created_via_cmd.yml
patel-bhavin: Will remove "wscript.exe","cscript.exe" from SPL and keep the search and descirption the same

patel-bhavin:
Thanks for the comments and suggestions @nasbench - i have pushed the fixes and can you have one more look?

nasbench (approved):
Good work. LGTM.

I applied a final SPL beautify no logic change.

PR #3974, opened 2026-03-26
Update Suricata TA and Related Analytics

nasbench:
The old Suricata TA was nuked out from Splunkbase for "unknown reasons". We've identified a replacement TA that does the same parsing (overall). This PR aims to use that TA and update corresponding DS and `contentctl.yml` files. Below is the summary of the changes

### Summary

Because the new TA has a different parsing for the URL field (and a bunch of other fields) we had to make some changes. The CCX TA uses an eval to get the value of the URL CIM field, as follow

```yml
EVAL-url = if(
    isnotnull('http.url')
    AND
    match('http.url',"^(?:https?\:\/\/)?([^\/\:]+).*$"),'http.url',
    
    if(isnotnull('http.url')
    AND
    isnotnull('http.hostname'),
    tostring('http.hostname')+tostring('http.url'),
    
    if(isnotnull('http.url')
    AND
    isnull('http.hostname'),
    'http.url',null())))
```

Which means the URL is usually a combination of `http.url` and `http.hostname`.

In contrast, the old TA was simply doing an alias to URL form http.url (which is technically incorrect). Since `http.url` actually contains the URI.

In practice, this means that we have to add a wildcard at the start of most of the analytic using the URL logic. Since it will be prefixed with the hostname. Most of the time this will not be an issue as long as the anchor string is long. I've only identified a couple of analytics suffering from small anchors and have proposed a fix. See details below.

### Updated Analytics

As explained above most analytics have seen an addition of a wildcard character at the start of the string. I also took the time to update their metadata when needed and beautified the YAML/SPL. I will only detail meaningful fixes below.

- `Disable Schedule Task` / `Wscript Or Cscript Suspicious Child Process` both of these analytics we demoted to `Anomaly` but never saw their score reduced. Hence I've updated them here.
- `Adobe ColdFusion Access Control Bypass` - Update the logic to remove non double slash URLs. Since according to the reference they are not needed.
- `HTTP Request to Reserved Name on IIS Server` - This analytic suffered from a short anchor logic. Hence I've updated the search to reduce FP by adding an extra check after the tstats.

### Deprecated Analytics

- `Ivanti Sentry Authentication Bypass` - After checking the reference, there is no strong IOC to catch this behavior and the actual analytic is weak. Hence it was deprecated.

---

> Note: As a side bonus in theory all of these changes will work with the old TA if someone still has it installed.


nasbench:
Marking as draft, until i get the fixes deployed.

ljstella (approved):
LGTM

patel-bhavin (approved):

PR #3983, opened 2026-03-31
Vmware tools updates

RavenTait:
Add some filtering to the vmware tools child process detection


detections/endpoint/windows_suspicious_vmware_tools_child_process.yml
```diff
@@ -15,6 +15,8 @@ search: |-
       WHERE Processes.parent_process_name=vmtoolsd.exe
         AND
         Processes.process_name IN ("powershell.exe","cmd.exe", "msbuild.exe", "microsoft.workflow.compiler.exe", "searchprotocolhost.exe", "scrcons.exe", "cscript.exe", "wscript.exe","bitsadmin.exe", "rundll32.exe", "wmic.exe", "mshta.exe", "certutil.exe", "schtasks.exe")
+        AND NOT
+        Processes.process IN ("*\\VMware Tools\\poweroff-vm-default.bat*","*\\VMware Tools\\poweron-vm-default.bat*")
```
nasbench: To avoid FNs. Can we add `cmd.exe` as a `process_name` in combination to this filter ?

nasbench (approved):

PR #3991, opened 2026-04-02
Group by actual template names instead of file name

henryy-splunk:
### Details
Previously we are grouping the response templates by file name. This means that if the file name gets changed SelfReplicated_v1.json -> Self-Replicated_v2.json, it would be a new line in the manifest for response template.

The change here is to group them by the actual response template name, meaning if the response template file is 
SelfReplicated_v1.json -> name: Self-Replicated, version: 1
Self-Replicated_v2.json -> name: Self-Replicated, version: 2
they would be grouped as the same.

Also add validation such that we have unique response_template name and version pairs.

Things to look out for:
The response template file names and links are url encoded.

### Checklist

- [ ] Validated SPL logic.


kbouchardherjavecgroup (commented):
thank you.

henryy-splunk:
> thank you.

Can I get your approval as well? Thanks!

kbouchardherjavecgroup (dismissed):
looks good.

henryy-splunk:
Please merge in after https://github.com/splunk/security_content/pull/3990 gets merged. Thank you

pyth0n1c (dismissed):

.github/workflows/response_templates/template_script.py
```diff
@@ -67,54 +70,67 @@ def _get_template_mapping(directory):
     path = Path(directory)
     if not path.exists() or not path.is_dir():
         raise ValueError(f"The directory {directory} does not exist or is not a directory.")
-    
-    # Check for non-JSON files
-    non_json_files = [f.name for f in path.iterdir() if f.is_file() and f.suffix != '.json']
+
+    # Check for non-JSON files (ignore hidden files like .DS_Store)
+    non_json_files = [f.name for f in path.iterdir() if f.is_file() and f.suffix != '.json' and not f.name.startswith('.')]
```
pyth0n1c: We should not ignore these files. Only json files should be allowed - unless there is an intentional use case for this, non json files should be _explicitly_ forbidden.
As far as I am aware, there is no reason that a hidden file like .DS_Store should be added to the git repo. The only one I can think of as legitimate is `.gitkeep` which, if applicable, should be a very explicit inclusion.

pyth0n1c (commented):
This change is not required, but recommended.

.github/workflows/response_templates/template_script.py
henryy-splunk: Updated to continue to forbid . files.

ljstella (approved):
:shipit: i guess

Issue #4000, opened 2026-04-08
[BUG] Missing fields for Detection: Windows Event Log Cleared


AndreiBanaru:
For the [Windows Event Log Cleared](https://github.com/splunk/security_content/blob/v5.25.0/detections/endpoint/windows_event_log_cleared.yml#L17-L19) detection, some of the fields won't be available, and some results will be missed.

I'm refering to:
- app
- name
- object_attrs
- signature
- src_user
- subject

which don't appear when `source=XmlWinEventLog:System EventCode=104` .

I've checked the Splunk_TA_windows add-on and at least for `name`, `signature `and `subject `these come from the **windows_signatures_910.csv** doesn't have the `104 `signature_id in it.

The app field comes from the **windows_apps.csv** as an enrichment and there is no association for it to be populated since neither `EventCode=104` or  `source=XmlWinEventLog:System` are present.

I'm not an expert in attack_data but I believe tests passed because the [data ](https://github.com/splunk/attack_data/blob/master/datasets/attack_techniques/T1070.001/windows_event_log_cleared/windows-xml.log) has both `XmlWinEventLog:System` and `XmlWinEventLog:Security` events in it, and it gives a result for the `XmlWinEventLog:Security`, which the tests consider sufficient.

patel-bhavin:
The above PR has been merged! thank you @AndreiBanaru for contributing these fixes!

PR #4001, opened 2026-04-08
Remove missing fields in Windows Event Log Cleared detection

AndreiBanaru:
As described in #4000, I believe these fields should be removed since the aggregation will miss results for `EventCode=104`.

### DIFF
--- a/data_sources/windows_event_log_security_1102.yml
+++ b/data_sources/windows_event_log_security_1102.yml
@@ -1,7 +1,7 @@
 name: Windows Event Log Security 1102
 id: 8db7b91a-6d7a-40e7-bfac-06f8e901a9cb
-version: 3
-date: '2025-07-10'
+version: 4
+date: '2026-04-15'
 author: Patrick Bareiss, Splunk
 description: Logs an event when the audit log is cleared.
 mitre_components:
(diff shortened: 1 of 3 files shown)


AndreiBanaru:
Bummer, I see the build failed since `Windows Event Log Security 1102` data source expects these fields:
> Value error, Data source 'Windows Event Log Security 1102' has output fields ['app', 'name', 'object_attrs', 'src_user', 'subject'] that are not present in the search for detection 'Windows Event Log Cleared'

AndreiBanaru:
I could do a:

```
| fillnull app name object_attrs signature src_user subject value=unknown
```

right before the `stats `, but in my opinion maybe you guys would want to reconsider updating contentctl to just throw a warning if the `output_fields `aren't all present in the query.

nasbench:
> Bummer, I see the build failed since `Windows Event Log Security 1102` data source expects these fields:
> 
> > Value error, Data source 'Windows Event Log Security 1102' has output fields ['app', 'name', 'object_attrs', 'src_user', 'subject'] that are not present in the search for detection 'Windows Event Log Cleared'

This is not an issue, we could remove these from the required fields, if they are not filled. I will investigate and apply the changes accordingly.

nasbench (approved):
LGTM

patel-bhavin (approved):

PR #4011, opened 2026-04-13
Improved detections based on telemetry data

P4T12ICK:
Improved detections based on telemetry data


detections/endpoint/possible_lateral_movement_powershell_spawn.yml
```diff
@@ -1,10 +1,10 @@
 name: Possible Lateral Movement PowerShell Spawn
 id: cb909b3e-512b-11ec-aa31-3e22fbd008af
-version: 14
-date: '2026-03-26'
+version: 15
+date: '2026-04-13'
 author: Mauricio Velazco, Michael Haag, Splunk
 status: production
-type: TTP
+type: Anomaly
```
nasbench, line 7: Did you want to keep the score 50 for this?

detections/endpoint/windows_msiexec_spawn_discovery_command.yml
```diff
@@ -1,18 +1,18 @@
...
 data_source:
     - Sysmon EventID 1
     - Windows Event Log Security 4688
     - CrowdStrike ProcessRollup2
 search: |-
     | tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime FROM datamodel=Endpoint.Processes
-      WHERE Processes.parent_process_name=msiexec.exe Processes.process_name IN ("powershell.exe", "pwsh.exe","cmd.exe", "nltest.exe","ipconfig.exe","systeminfo.exe")
+      WHERE Processes.parent_process_name=msiexec.exe Processes.process_name IN ("powershell.exe", "pwsh.exe", "nltest.exe","ipconfig.exe","systeminfo.exe")
```
nasbench, line 15: Can you add reasoning for the filter, so that people do not add it in the future without consideration.

detections/endpoint/windows_new_deny_permission_set_on_service_sd_via_sc_exe.yml
```diff
@@ -43,6 +43,7 @@ search: |-
         "*;LS*",
         "*;S-1-5-19*"
     )
+    NOT Processes.process IN ("*McAfeeFramework*", "*mfefire*", "*mfemms*", "*mfevtp*", "*macmnsvc*", "*masvc*")
```
nasbench, line 46: Please add a note in the rule to make people aware about this filter. So for those who do not use McAfee products make the necessary changes if they want

detections/endpoint/regsvr32_silent_and_install_param_dll_loading.yml
```diff
@@ -14,7 +14,7 @@ search: |-
     | tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime FROM datamodel=Endpoint.Processes
       WHERE `process_regsvr32`
         AND
-        Processes.process="*/i*"
+        Processes.process="*/i*" AND NOT Processes.process="*Microsoft\\TeamsMeetingAddin*"
```
nasbench, line 17: Please add a note in the FP section about this and teams

nasbench:
@P4T12ICK if you address these we can ship this in 5.26 next week

patel-bhavin:
moving this to 5.27.0!

detections/endpoint/windows_msiexec_spawn_discovery_command.yml
nasbench: I improved the detection itself since it was looking for discovery commands. It was incorrect. I added specific children and improved the logic.
Check in telemetry and with the new query we got good results :D

detections/endpoint/regsvr32_silent_and_install_param_dll_loading.yml
nasbench: `regsvr32_silent_and_install_param_dll_loading.yml`  and `regsvr32_with_known_silent_switch_cmdline.yml`. I have pinged teo, and added a comment about teams

nasbench (approved):

Issue #3976, opened 2026-03-27
[BUG] Broken drilldown: "View risk events for the last 7 days for ..."


AndreiBanaru:
With the release of Enterprise Security 8, the risk drill down searches for TTP type detections are broken.

I'm referring to [these](https://github.com/splunk/security_content/blob/develop/detections/cloud/asl_aws_credential_access_getpassworddata.yml#L34-L36):

```
drilldown_searches:
    - name: View risk events for the last 7 days for - "$user$"
      search: '| from datamodel Risk.All_Risk | search normalized_risk_object IN ("$user$") starthoursago=168  | stats count min(_time) as firstTime max(_time) as lastTime values(search_name) as "Search Name" values(risk_message) as "Risk Message" values(analyticstories) as "Analytic Stories" values(annotations._all) as "Annotations" values(annotations.mitre_attack.mitre_tactic) as "ATT&CK Tactics" by normalized_risk_object | `security_content_ctime(firstTime)` | `security_content_ctime(lastTime)`'
      earliest_offset: $info_min_time$
      latest_offset: $info_max_time$
```

I'm pinning this to a change done in the `SA-ThreatIntelligence` add-on. The **baseSearch** in the **Risk.json** datamodel now calls out an ``spath_annotations`` macro at the end. I believe this doesn't allow the use of `starthoursago` parameters later in the SPL.

What could potentially solve the problem would be:
1. Removal of `starthoursago=168` from the search key
2. Setting `earliest_offset=-7d` and `latest_offset=now`

This would technically achieve the same outcome as before, but I wasn't able to find the right way to make the `-7d` and `now` expand.

patel-bhavin:
Thanks for reporting this issue : We will have a look at this!

patel-bhavin:
Also, confirmed that starthoursago=168 no longer works and I have tested this fix : 
and - 

### Testing -7d and now - gives a UI validation error 

<img width="300" height="300" alt="Image" src="https://github.com/user-attachments/assets/21946633-b0f0-4d9a-a0a6-abdc90ee3c28" />

### Updated to use  7d and 0 as the config and removed starthoursago from SPL:

<img width="300" height="300" alt="Image" src="https://github.com/user-attachments/assets/1dda5d38-7f21-4bdf-a8d9-cc54bb868bf3" />

We can get this update shipped out in the 5.26.0 release. Thanks for this awesome feedback @AndreiBanaru !

patel-bhavin:
[...]

PR #4016, opened 2026-04-15
updating risk drilldowns

patel-bhavin:
Fixes for issue : https://github.com/splunk/security_content/issues/3976


nasbench (dismissed):
Did not look at all for my sanity. But since this a search/replace it should all be good. I'll take care of the versions

AndreiBanaru:
This would show results but I think it doesn't render the intention:
```
  earliest_offset: 7d
  latest_offset: "0"
```
I've updated it for 1 detection in our dev environment and this is what I see in Splunk Web:
image

and when I click on the drilldown for a finding I have:
image

patel-bhavin:
Hello @AndreiBanaru : i installed a build of the ESCU from this specific PR onto our test instance and it seems like this works as intended : You can note the time on the right side of the screenshot and i recommend to look under 
"date & Time Range" to confirm if the time computation works as expected! 

What splunk and ES version are you on ?  I have tested this ES 8.4 and Splunk 10

I also noticed that ES adds the `y` and renders it in UI as `0y` for latest_offset however this seems like a bug in ES and the functionality works as expected! Based on our testing : we will likely move forward with getting this PR shipped! Let us know if you have any concerns

image

### Note:

- Just FYI, There are a bunch of issues around drilldowns in ES 8.1 : https://help.splunk.com/en/splunk-enterprise-security-8/release-notes-and-resources/8.1/splunk-enterprise-security-release-notes/known-issues

nasbench (approved):

AndreiBanaru:
Hi @patel-bhavin , I've tested with ES 8.4 and Splunk 9.4.8 yesterday.
Today I've upgraded to 10.2.1 and I noticed the same behavior.

I'll keep my concerns to myself for now, but `0` usually means **January 1, 1970 at 12:00:00 AM** as per [docs](https://help.splunk.com/en/splunk-enterprise/spl-search-reference/9.1/time-format-variables-and-modifiers/time-modifiers#ariaid-title3).

For me, "Data and Time Range" looks different:
image

patel-bhavin:
@AndreiBanaru - this is really odd : While 0 usually means January 1, 1970 at 12:00:00 AM , it is not entirely true for the way offsets are working in ES as how I understand it : [drill downs docs](https://help.splunk.com/en/splunk-enterprise-security-8/administer/8.4/investigations/investigate-findings-using-drilldown-searches-and-dashboards-in-splunk-enterprise-security) : 
image



I also ran this on Splunk 10.2 and ES 8.4 today and we are seeing expected results as far as time modifiers are concerned. 

image

and similar is the case with Earliest and Latest values in epoch 

image

patel-bhavin:
@AndreiBanaru - we will ship this update for 5.26.0 however, if this solution is still concerning please let us know! Maybe I am missing something here but we have tested this on 2 splunk versions and created identical results . Thank you again for being awesome and keeping ESCU on a check beyond these CI systems!

PR #4036, opened 2026-04-28
Mitre updates

ljstella:
### Details

Changes out the `external_repo/cti` clone for the STIX2.1 version, which functions with the version of `attackcti` that we're using in contentctl (version pinned there due to pydantic). 

Will add the MITRE updates required to build to this repo once we can get the CI running with the new version of contentctl

Associated contentctl changes: https://github.com/splunk/contentctl/pull/479

This also currently includes the changes in #4035 

### Checklist

- [ ] Validated SPL logic.


ljstella:
Migrated the techniques per https://medium.com/mitre-attack/attack-v19-ff329cb65d66 and the Crosswalk. Held off on version bumps to make sure I wasn't double bumping.

pyth0n1c:
Probably obvious, but all of these detections will need a bump to their `version` + `date` fields.

nasbench:
Leaving this PR last before release.

ljstella:
Still have a couple lookups to add but appinspect should be passing now

.github/workflows/appinspect.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           echo "- Contentctl version - $(cat requirements.txt)"
           pip install -r requirements.txt
           git clone --depth=1 --single-branch --branch=master https://github.com/redcanaryco/atomic-red-team.git external_repos/atomic-red-team
-          git clone --depth=1 --single-branch --branch="ATT&CK-v18.1" https://github.com/mitre/cti external_repos/cti
```
patel-bhavin, line 25: Should we pin it to the latest version tag?

https://github.com/mitre-attack/attack-stix-data/tree/v19.0

ljstella: Because of the way that the STIX2.1 repo is organized with a single file per version, the "what version do we use?" part is controlled on the `contentctl` side where we select the file. In the current tooling its hardcoded to v19.0, but I've made notes about shifting that value into the configuration instead- we'll still be "pinned" to a version and can do the migration ourselves by hand rather than the old system where we found out the new version was released when CI broke and then we all scrambled to get the fixes in so we could all be unblocked.

ljstella:
**Status of lookups:**

- lookups\suspicious_named_pipes.csv

Status: There's overlap but only partial overlap with Mitre Tactic names here, probably fine to leave the "Defense Evasion" ones as is.

- lookups\windows_suspicious_tasks.csv

Status: There's overlap but only partial overlap with Mitre Tactic names here, probably fine to leave the "Defense Evasion" ones as is.

- lookups\windows_suspicious_services.csv

Status: There's overlap but only partial overlap with Mitre Tactic names here, probably fine to leave the "Defense Evasion" ones as is.

- app_template\lookups\mitre_enrichment.csv

Status: Done, added in previous commit

- lookups\hijacklibs_loaded.csv

Status: No changes needed, included techniques were not migrated




That being said, this should now be good to go.

nasbench (approved):
LGTM

PR #4081, opened 2026-05-12
Copyfail and DIrtyFrag

RavenTait:
### Details

Two new detections around Copy Fail and one new data source for Linux kern.log

Linux Malformed Auth Entry
Linux PF_ALG Registration Outside of Boot Window


nasbench (changes requested):
This PR is using the old format, can you please update it to use the new one.

detections/endpoint/linux_su_launch_process_with_null_argv.yml
```diff
@@ -0,0 +1,68 @@
+name: Linux Su Launch Process with Null Argv
```
nasbench: The detection logic focus on any process not just `su` so lets update the title/filter macro and any related metadata to reflect this.

detections/endpoint/linux_malformed_auth_entry.yml
```diff
@@ -0,0 +1,74 @@
+name: Linux Malformed Auth Entry
+id: 962602e9-08ae-47ff-af89-a1e83f461a26
+version: 1
+creation_date: '2026-05-06'
+modification_date: '2026-05-06'
+author: Raven Tait, Splunk
+status: production
+type: TTP
```
nasbench: I think this should be demoted to an anomaly instead. Since this triggering is not investigation worthy yet.

data_sources/linux_messages_syslog.yml
```diff
@@ -0,0 +1,56 @@
...
+- tag
+- tag::action
+- tag::eventtype
+- timeendpos
+- timestartpos
+- user
+- user_name
+- vendor_action
+- vendor_product
+example_log: 'May  6 17:24:00 ubuntu kernel: [  801.664268] NET: Registered PF_ALG protocol family'
```
nasbench:
```suggestion
example_log: 'May  6 17:24:00 ubuntu kernel: [  801.664268] NET: Registered PF_ALG protocol family'

```

data_sources/linux_messages_syslog.yml
```diff
@@ -0,0 +1,56 @@
+name: Linux Messages Syslog
+id: c9e3bbb5-e0a2-4bac-8457-227e71841bda
+version: 2
```
nasbench:
```suggestion
version: 1
```

nasbench (approved):

patel-bhavin (approved):

PR #4085, opened 2026-05-18
Workflow updates

pyth0n1c:
Includes code snippets for generating new content.
To use them:

1. Create a new yml file in the relevant directory. For instances, create detections/endpoint/new_detection.yml

2.  Open the file.

3. Start typing detection.  It will autopopulate with the snippet `detection_escu6` and hit ENTER to autopopulate the file with the contents of the Snippet.

This works for all types of content, including detections, baselines, stories, macros,lookups, dashboards, and data_sources. It does not work with schedules as those are not a commonly created type of content.

This also includes workflow updates to build, inspect, and unit test content. There are issues with these workflows at this time:

1. unit-testing.yml is failing at the moment because it relies on testing workflows being present in the target branch (for installing contentctl-ng, which they are not yet).
2. Build fails because there are still a large number of pieces of content which require MANUAL REVIEW updates
3. Appinspect fail for the same reason as above - build is failing.


pyth0n1c:
Closing this PR out, as we are taking a different approach for this update.

PR #4088, opened 2026-05-19
ESCU 6 Manual Migrations

ljstella:
### Details

Working through rba_upgrade_tracking.json

done so far: 

- "Correlation" type searches that did not have entities
- "Multiple non-user entities but no user entities"
- Unbalanced $ in risk message
- Multiple user-type entities
- Messages missing a $ or not referencing a field
- Baselines (see below)


ljstella:
Baselines: 

baselines/dnstwist_domain_names.yml <- flagged because it previously listed a detection that used it, but that detection doesn't exist. However, there are other detections that use it, so it has to stay for now.

 modified:   baselines/baseline_of_network_acl_activity_by_arn.yml <- dangling
 modified:   baselines/baseline_of_security_group_activity_by_arn.yml <- dangling
 modified:   baselines/create_a_list_of_approved_aws_service_accounts.yml <- dangling
 modified:   baselines/discover_dns_records.yml <- dangling
 modified:   baselines/previously_seen_command_line_arguments.yml <- dangling
 
As far as I can tell, these others are "dangling" - they are no longer referenced by any existing detections. Not easy to tell if other folks are still using them for their own purposes, but I'd suggest we follow up post 6.0 launch and mark these for removal.

ljstella:
`contentctl-ng build` now successfully builds a package!

detections/cloud/o365_bec_email_hiding_rule_created.yml
```diff
@@ -40,11 +40,11 @@ finding:
         type: user
         score: 50
 threat_objects:
-    - &id001
-      field: Name
+    - field: Name
       type: signature
 analytic_story:
     - Office 365 Account Takeover
+data_source: []
```
nasbench: This should be moved up. Except if we have an ordering script that is going to be applied later.

detections/web/monitor_web_traffic_for_brand_abuse.yml
```diff
@@ -23,7 +23,7 @@ how_to_implement: You need to ingest data from your web traffic. This can be acc
 known_false_positives: No false positives have been identified at this time.
 references: []
 finding:
-    title: Potential Brand Abus discovered in web logs
+    title: Potential Brand Abus discovered in web logs from $src$
```
nasbench:
```suggestion
    title: Potential brand abuse discovered in web logs from $src$
```

detections/application/okta_risk_threshold_exceeded.yml
```diff
@@ -51,6 +51,9 @@ tests:
           source: risk_data
           sourcetype: stash
       test_type: unit
-MANUAL_REVIEW:
-    rba: {}
-    manual_review_rationale: Legacy Correlation detections have no rba section (and therefore no entities), but the new format requires a finding with at least one entity. A content author must supply the finding entity for each Correlation detection. Additionally, evaluate whether any Threat Objects are appropriate.
+finding:
```
patel-bhavin: can we re arrange the finding key such that its consistent with the rest of the yamls?  also, it seems rather incorrect to have to set score of 0 of findings

detections/cloud/aws_s3_exfiltration_behavior_identified.yml
```diff
@@ -53,6 +53,9 @@ tests:
           sourcetype: stash
           source: aws_exfil
       test_type: unit
-MANUAL_REVIEW:
-    rba: {}
-    manual_review_rationale: Legacy Correlation detections have no rba section (and therefore no entities), but the new format requires a finding with at least one entity. A content author must supply the finding entity for each Correlation detection. Additionally, evaluate whether any Threat Objects are appropriate.
+finding:
```
patel-bhavin: can we re arrange the finding key such that its consistent with the rest of the yamls?

detections/application/okta_risk_threshold_exceeded.yml
ljstella: This is _mostly_ maintaining the existing behavior of those searches - we generally don't want to create even more risk events/add to the score of things we're highlighting specifically because of other risk events, lest we create a feedback loop where the Correlation search triggering causes the next run of that same search to trigger again.

detections/cloud/azure_ad_privileged_role_assigned.yml
```diff
@@ -46,12 +46,12 @@ drilldown_searches:
 finding:
     title: A privileged Azure AD role was assigned for User $user$ initiated by $initiatedBy$
     entity:
-        field: user
+        field: initiatedBy
         type: user
```
patel-bhavin, line 50: @ljstella @pyth0n1c - findings do not have a score when operating via the UI

detections/application/okta_risk_threshold_exceeded.yml
pyth0n1c: Per Enterprise Security 8 (and some content currently in develop today), Score _CAN_ be 0:
https://help.splunk.com/en/splunk-enterprise-security-8/administer/8.0/risk-based-alerting/risk-scoring-in-splunk-enterprise-security#:~:text=Colors%20are%20used%20to%20distinguish%20between%20the%20levels%20of%20risk.%20A%20yellow%20badge%20represents%20a%20risk%20score%20of%200%2D25%2C%20orange%20represents%2025%2D50%2C%20light%20red%20represents%2050%2D75%2C%20and%20dark%20red%20represents%20a%20risk%20score%20above%2075.

This is what we validate against in CICD today.
At your request, we _can_ make this strictly greater than 0, as you are welcome to be more strict on your content than ES8 allows.
I defer to discussion from you and Lou (and the rest of the team) if this is a tooling change you would like :)

detections/cloud/o365_email_access_by_security_administrator.yml
```diff
@@ -36,12 +36,12 @@ drilldown_searches:
 finding:
     title: A security administrator $src_user$ accessed email messages for $user$
     entity:
-        field: user
+        field: src_user
         type: user
         score: 50
```
patel-bhavin, line 41: this finding has a score of 50 as opposed to zero. I think we should remove score for findings?

detections/cloud/aws_s3_exfiltration_behavior_identified.yml
ljstella: Keys have been re-ordered throughout

detections/cloud/azure_ad_privileged_role_assigned.yml
ljstella: Not sure what you mean here?

pyth0n1c: @patel-bhavin can you please go into some more detail about this?
Please see the following screenshots, which show that entities associated with findings do have a score associated with each entity.

ES 8.4:

Image

ES 8.5:

Image

patel-bhavin: my bad! ignore this comment!

detections/cloud/o365_email_access_by_security_administrator.yml
ljstella: The score on the Finding is because we're no longer creating separate duplicate intermediate findings with the risk - So scores of 50/20 migrated over based on TTP/Anomaly mappings previously done. If we put a score of 0 on the Finding, we'd create a risk event with a 0 score for the thing in the entity, and then only create scored risk events for the intermediate finding entities.

detections/application/okta_risk_threshold_exceeded.yml
patel-bhavin: cool! these are specifically for type: Correlation! 

Looks good!

pyth0n1c (approved):
Thanks Lou! All these changes look great!
This has my approval from the SCE Side!

I have also committed updated schemas/ myself. These were updated since the MANUAL_REVIEW content now parses correctly, which means they make it into the enums that make up the schema files.

PR #4112, opened 2026-06-01
Fix case-sensitive AD self-add group comparison

### LINKED ISSUE #4105: [BUG]Logic Problem of Windows AD add Self to Group
Hello Splunk Security Team,
i was auditing this scenario after that didnt see any output in my test suit auditing, i checked the scenario and find out that in the scenario logic line:
`| where user=src_user`
the value of user and src_user are like this user=DA-1 and src_user=da-1 and the scenario wont fire.
so i changed this line to this:
`| where lower(user)=lower(src_user)`
and the problem fixed 
please modify the scenario logic
regards,
Mahdi Hamedani Nezhad

srkyn:
## Summary

Fixes the `Windows AD add Self to Group` analytic so the self-add comparison is case-insensitive:

```spl
| where lower(user)=lower(src_user)
```

## Why

Issue #4105 reports that matching can fail when the same account appears with different casing, for example `user=DA-1` and `src_user=da-1`.

## Validation

```bash
python scripts/validate_yaml.py detections/endpoint/windows_ad_add_self_to_group.yml
```

Result:

```text
[PASS] All 1 file(s) passed validation!
```

Fixes #4105.


srkyn:
Rechecked this branch after the failed CI run. The PR diff is still limited to the intended case-insensitive comparison plus version/date bump, and the detection YAML parses cleanly.

Local sanity check passed:
- confirmed lower(user)=lower(src_user) is present
- confirmed version is bumped to 11
- parsed detections/endpoint/windows_ad_add_self_to_group.yml with PyYAML

The two failing GitHub Actions jobs appear to be blocked before exercising this detection because forked PR runs do not receive Splunk credentials:
- appinspect: APPINSPECTUSERNAME / APPINSPECTPASSWORD are empty, then Splunk API login returns 400 Client Error
- unit-testing: contentctl-ng install exits with Splunkbase credentials not provided in environment variables

The other checks (YAML Validation, build, and response template build) passed.

patel-bhavin:
@srkyn - thank you for the PR and doing the testing locally, I think the fix makes sense and I will raise this issue of unit-testing needing creds. We recently updated our tooling and didnt think of this issue coming up! I will get back to you!

srkyn:
Thanks, that makes sense. I will leave the branch as-is unless you want a change from my side. The detection diff is intentionally small: case-insensitive comparison only, plus version/date metadata.

nasbench (approved):

PR #4122, opened 2026-06-11
Add Dirty Frag (CVE-2026-43284/43500) Linux kernel LPE detection

Axselll:
Adds detection for Dirty Frag, a deterministic Linux kernel local privilege escalation disclosed May 7 2026 and exploited in the wild within 24 hours. Attaches to the existing Linux Privilege Escalation analytic story.

Dirty Frag chains two kernel page-cache write vulnerabilities in the xfrm-ESP (CVE-2026-43284) and RxRPC (CVE-2026-43500) subsystems. It bypasses the Copy Fail mitigation (algif_aead blocklist) — environments that patched for Copy Fail remain fully exposed.

Detection uses auditd session-scoped auid correlation to identify the universal exploitation pattern: ELF binary execution from a user-writable path followed by a confirmed privilege transition (euid=0, uid!=0) within the same audit session within 5 minutes. Validated against the public PoC (V4bel/dirtyfrag) in a sandboxed lab environment.

Attack data: splunk/attack_data#1182

### Details

### Checklist

- [ ] Validated SPL logic.


Axselll:
Screenshot from 2026-06-12 02-05-46

Tested on my own lab

detections/endpoint/linux_dirty_frag_kernel_privilege_escalation.yml
```diff
@@ -0,0 +1,116 @@
...
+  | stats min(eval(if(is_elf_stage=1, _time, null()))) AS elf_time min(eval(if(is_lpe=1, _time, null()))) AS lpe_time values(exe) AS executables values(comm) AS commands values(uid) AS uids values(euid) AS euids BY auid, host, UID 
+  | where isnotnull(elf_time) AND isnotnull(lpe_time) 
+  | where lpe_time > elf_time 
+  | eval window_secs = lpe_time - elf_time 
+  | where window_secs <= 300 
+  | `security_content_ctime(elf_time)`
+  | `security_content_ctime(lpe_time)`
+  | rename host AS dest UID AS user
+  | table dest, auid, elf_time, lpe_time, window_secs, executables, commands, uids, euids, user 
+  | `linux_dirty_frag_kernel_privilege_escalation_filter`
```
nasbench: This currently has a couple of issues.

- First the search is not looking for DirtyFlag activity instead it is simply searching for exec event of binaries from some directories and then specifically focusing on SU execution, which how Su/Sudo will look like in normal operation AKA uid != euid (due to priv change).
- The search is also too broad at the start, you'll have to filter early or else this will take a longer range of time to complete since you are grabbing ALL syscalls and then filtering.

We already have a couple of detections covering dirtyflag https://github.com/splunk/security_content/pull/4081

If you want to keep this PR, you'll have to add other indicators of dirty flags and make it more specific. Adding things like SYSCALL="splice" OR SYSCALL="sendto" for example, or other function calls from the KERN_MODULE syscall...etc.

I will leave it up to you to choose the best indicators.

nasbench (changes requested):

detections/endpoint/linux_dirty_frag_kernel_privilege_escalation.yml
Axselll: let me see what i can do, thanks

Axselll: Hello @nasbench,

Understand that on my previous PR submission was correctly identified as a generic privilege-escalation pattern that did not anchor on Dirty Frag-specific behavior. After reviewing and tinkering around for a while i have came up with 3 points;

1. The previous version triggered on euid=0 AND uid!=0 combined with an ELF execution from a user-writable path. As you noted, this is the signature of any su/sudo invocation following any home-directory binary execution — not Dirty Frag. The new detection anchors on the splice() syscall volume signature that is unique to the exploit's page-cache spraying mechanism. The V4bel public proof-of-concept produces approximately 460 splice() calls within 90 seconds from a single unprivileged process. This volume from a user-writable path is not produced by normal user workloads, su, sudo, or service binaries.

2. The previous version pulled all type=SYSCALL events and filtered with eval. The new base search filters by key=dirty_frag_splice at the index, scoping events to the splice syscalls captured by the deployed audit rule. Performance is equivalent to other ESCU Linux auditd detections.

3. From my understanding and observation the existing detections under #4081 cover priv-esc artifacts and post-exploitation behavior. They do not anchor on the splice spray signature that is the kernel-mode action of the exploit itself. This detection covers the gap by identifying the exploitation event before privilege transition completes, providing earlier detection in the attack chain.

This detection requires two non-default auditd rules (dirty_frag_splice & dirty_frag_privesc) which make me confuse on where should deploy these dependency, other thing that i got from tinkering around is multiple detections on released version of ESCU require specific audit rules to be deployed for the analytic to function. Examples include detections relying on:

- key=auditd_add_user events
- Watches on /etc/passwd, /etc/shadow, /etc/sudoers
- Specific syscall captures for process injection or module loading

Last but not least, i already test everything in my lab. Please refer to my screenshot below;

testtt

Kindly let me know if you are OK with this and we can move on to the deployment part especially the custom audit rules

Thanks.

nasbench: Thanks for the update i will check it out later/early next week.

To answer your question, the config usually is this one https://github.com/Neo23x0/auditd/

But often you will have to mention this in the how to implement section, that is how the user will be aware of any conditions/updates needed.

If you have the necessary config please contribute it to that repo as well as paste it in the how to implement section.

As far as the other PR is concerned, I am aware of the coverage it gives and why i wanted to keep this one as well.

Axselll: Hello,

Just wants to let you know i have adjust the detection on this repository and also adjust some stuff on attack data repository, please also check that out.

I have been digging around on the auditd repository that you mention and found out that we can utilize just the splice_user and process_creation, other thing i found is direct telemetry on the xfrm-ESP and RxRPC kernel subsystems is not available via auditd syscall captures. The vulnerable code paths are reached through netlink message contents (xfrm-ESP) and indirect kernel call chains (RxRPC) that auditd cannot inspect. The splice() syscall is the kernel-mode action consistently captureable across exploit variants and is the anchor for this detection. From my understanding i don't think people will ingest kernel subsystem log just to cover this up, so splice_user and process_creation is our best shot. 

I already tested it on my environment as you can see the screenshot below;

image

Happy weekend.

nasbench (commented):
The overall new logic LGTM. The usage of the join can lead to edge cases where no results are returned since there is an inherit limit of 50K events, and since the events in questions are syscalls of process_creation, the limit could be surpassed easily in certain env.

But this is a good enough compromise currently. Eventually this could be improved to use multiple `stats` with change to the logic.

I will approve this once the attack_data issue has been handled.

detections/endpoint/linux_dirty_frag_kernel_privilege_escalation.yml
```diff
@@ -0,0 +1,169 @@
...
+product:
+    - Splunk Enterprise
+    - Splunk Enterprise Security
+    - Splunk Cloud
+category: endpoint
+security_domain: endpoint
+tests:
+    - name: True Positive Test
+      attack_data:
+        - data: https://media.githubusercontent.com/media/splunk/attack_data/master/datasets/attack_techniques/T1068/dirty_frag/dirty_frag_events_trimmed.log
```
nasbench:
```suggestion
        - data: https://media.githubusercontent.com/media/splunk/attack_data/master/datasets/attack_techniques/T1068/linux_dirtyfrag/dirty_frag_events_trimmed.log
```

nasbench (dismissed):

nasbench:
image

patel-bhavin (approved):

patel-bhavin:
LGTM! This commit [d294f9b](https://github.com/splunk/security_content/pull/4122/commits/d294f9be4b433da377d171b850d5eed55881ba18) should fix the integration test errors

PR #4123, opened 2026-06-13
Update Windows EDRSilencer renamed binary detection coverage

raylee-hawkins:
## Summary

* Expands Windows EDRSilencer command-line matching so renamed binaries using `blockedr` can still be detected when `blockedr` is the final argument.
* Keeps the analytic on normalized process telemetry and avoids adding new data-source/schema entries.

## Validation

* YAML parse passed.
* Targeted EventBasedDetection schema validation passed.
* Targeted metadata/SPL/data_source sanity check passed.
* `git diff --check` passed.
* `contentctl validate` was previously attempted, but repo-wide validation is currently blocked by unrelated lookup/KVStore validation issues.

## Maintainer note

This keeps the change to one detection YAML and only narrows the existing command-line matching gap described in the issue.


nasbench:
This does not currently fully address the issue reported in 4101, which suggest adding a new rule covering EID 5447.

I moved this to 6.2 release for now, until we explore this furter.

nasbench (approved):

nasbench (commented):
Removed the link to the issue as this does not address it but it does add an enhancement of the CLI coverage. I also took the liberty to update the findings message and the TO

PR #4125, opened 2026-06-15
Rewrote HTTP Scripting Tool User Agent to Web DM

inspired:
This PR Rewrote the detection `HTTP Scripting Tool User Agent` to work with the Web DM instead of relying on raw logs.


detections/web/http_scripting_tool_user_agent.yml
```diff
@@ -10,18 +10,22 @@ description: This Splunk query analyzes web access logs to identify and categori
...
-           status
-      | `http_scripting_tool_user_agent_filter`
+    | tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime
+    from datamodel=Web
+    where nodename=Web.Proxy
+    by Web.http_user_agent Web.dest Web.src Web.status
+    | `security_content_ctime(firstTime)`
+    | `security_content_ctime(lastTime)`
+    | `drop_dm_object_name(Web)`
+    | lookup scripting_tools_user_agents tool_user_agent AS http_user_agent OUTPUT tool
```
nasbench, line 25: Let's add the eval to make sure the UA is always lowercased.
```suggestion
    | eval http_user_agent = lower(http_user_agent)
    | lookup scripting_tools_user_agents tool_user_agent AS http_user_agent OUTPUT tool
```

detections/web/http_scripting_tool_user_agent.yml
```diff
@@ -10,18 +10,22 @@ description: This Splunk query analyzes web access logs to identify and categori
...
+    | where isnotnull(tool)
+    | stats count
+        min(firstTime) as first_seen
+        max(lastTime) as last_seen
+        values(tool) as tool
+    by http_user_agent dest src status
+    | convert ctime(first_seen) ctime(last_seen)
```
nasbench: We use these macros as a convention in ESCU
```suggestion
    | `security_content_ctime(first_seen)`
    | `security_content_ctime(last_seen)`
```

detections/web/http_scripting_tool_user_agent.yml
```diff
@@ -10,18 +10,22 @@ description: This Splunk query analyzes web access logs to identify and categori
...
-      | where isnotnull(tool)
-      | rename dest_ip as dest
-      | stats count min(firstTime) as first_seen max(lastTime) as last_seen values(tool) as tool
-        BY http_user_agent dest src_ip
-           status
-      | `http_scripting_tool_user_agent_filter`
+    | tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime
+    from datamodel=Web
+    where nodename=Web.Proxy
+    by Web.http_user_agent Web.dest Web.src Web.status
```
nasbench: I know it was not in the original search. But let's also add a check to where http_user_agent is not null or empty just to gain a bit in performance and reduce the search space

detections/web/http_scripting_tool_user_agent.yml
```diff
@@ -45,10 +49,12 @@ intermediate_findings:
         - field: dest
           type: system
           score: 20
-          message: Known scripting tool was used against a web request. The source IP is $src_ip$ and the destination is $dest$.
+          message: Known scripting tool was used against a web request. The source IP is $src$ and the destination is $dest$.
```
nasbench: Just because src can be either an IP or Hostname (let's make that clear)
```suggestion
          message: Known scripting tool was used against a web request. The source IP/Host is $src$ and the destination is $dest$.
```

nasbench (commented):
Thanks for the updates, just a couple of small suggestions before we merge this

detections/web/http_scripting_tool_user_agent.yml
```diff
@@ -10,18 +10,24 @@ description: This Splunk query analyzes web access logs to identify and categori
...
-      | where isnotnull(tool)
-      | rename dest_ip as dest
-      | stats count min(firstTime) as first_seen max(lastTime) as last_seen values(tool) as tool
-        BY http_user_agent dest src_ip
-           status
-      | `http_scripting_tool_user_agent_filter`
+    | tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime
+    from datamodel=Web
+    where nodename=Web.Proxy
+    by Web.http_user_agent Web.dest Web.src Web.status
```
nasbench:
```suggestion
    from datamodel=Web where
    nodename=Web.Proxy
    Web.http_user_agent =*
    NOT Web.http_user_agent IN (
        "-",
        "unknown"
    )
    by Web.http_user_agent Web.dest Web.src Web.status
```

nasbench (changes requested):
~The only thing left to verify before merging this, is if the sourcetype in question is tagged to the web DM from the NGINX TA.~

~cc @RavenTait can double check when you have some time.~ (never mind, already did it)

detections/web/http_scripting_tool_user_agent.yml
```diff
@@ -10,18 +10,29 @@ description: This Splunk query analyzes web access logs to identify and categori
...
-      | lookup scripting_tools_user_agents tool_user_agent AS http_user_agent OUTPUT tool
-      | where isnotnull(tool)
-      | rename dest_ip as dest
-      | stats count min(firstTime) as first_seen max(lastTime) as last_seen values(tool) as tool
-        BY http_user_agent dest src_ip
-           status
-      | `http_scripting_tool_user_agent_filter`
+    | tstats `security_content_summariesonly` count min(_time) as firstTime max(_time) as lastTime
+    from datamodel=Web where
+    nodename=Web.Proxy
```
nasbench: After checking the TA. We have this definition

[eventtype=nginx_access]
web = enabled
inventory = enabled
activity = enabled

and 

[nginx_access]
search = sourcetype=nginx:plus:access OR sourcetype="nginx:plus:kv"

This means its tagged against the Web DM but not against proxy node. So we cannot use it

```suggestion
```

nasbench: FYI @inspired

nasbench (approved):

RavenTait (approved):

nasbench:
Tested and verified

PR #4134, opened 2026-06-23
Splunk Vuln content to Github

patel-bhavin:
moving from Gitlab to Github as these are public detections and listed on research.splunk.com


nasbench (approved):
The move LGTM. The searches however (a lot of them are questionable). I have deprecated the obvious ones that are no longer needed.

Left some that look okayish and are usable to some extent.

Also moved certain macros to the deprecated macro sub-folder since they are no longer used.

patel-bhavin:
:shipit:

PR #4138, opened 2026-06-26
explicitly list all apps in this repo's install.yml

pyth0n1c:
This is part of a migration to remove install.yml apps that are hardcoded in contentctl-ng's install.yml and move them into the scope of the security_content repo for clarity and control.

### DIFF
--- a/install.yml
+++ b/install.yml
@@ -25,9 +25,6 @@ apps:
   - appid: Splunk_TA_google-cloudplatform
     url: https://splunkbase.splunk.com/app/3088
     version: 5.0.3
-  - appid: Splunk_ML_Toolkit
-    url: https://splunkbase.splunk.com/app/2890
-    version: 5.7.4
   - appid: utbox
     url: https://splunkbase.splunk.com/app/2734
     version: 1.9.4
@@ -37,9 +34,6 @@ apps:
   - appid: TA-crowdstrike-falcon-event-streams
     url: https://splunkbase.splunk.com/app/5082
     version: 3.7.2
-  - appid: Splunk_SA_Scientific_Python_linux_x86_64
-    url: https://splunkbase.splunk.com/app/2882
-    version: 4.3.2
   - appid: SA-VMWIndex
     url: https://splunkbase.splunk.com/app/5640
     version: 4.0.4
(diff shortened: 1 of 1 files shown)


install.yml
```diff
@@ -1,3 +1,48 @@
...
+    version: 1.9.4
+  - appid: SA-admon
+    url: https://splunkbase.splunk.com/app/6853
+    version: 1.1.2
+  - appid: TA-crowdstrike-falcon-event-streams
+    url: https://splunkbase.splunk.com/app/5082
+    version: 3.7.2
+  - appid: Splunk_SA_Scientific_Python_linux_x86_64
+    url: https://splunkbase.splunk.com/app/2882
+    version: 4.3.2
```
nasbench: I think this is not needed since we removed the MLTK content form the repo.

install.yml
```diff
@@ -1,3 +1,48 @@
...
+    version: 4.1.5
+  - appid: Splunk_TA_aws-kinesis-firehose
+    url: https://splunkbase.splunk.com/app/3719
+    version: 1.3.2
+  - appid: Splunk_TA_google-cloudplatform
+    url: https://splunkbase.splunk.com/app/3088
+    version: 5.0.3
+  - appid: Splunk_ML_Toolkit
+    url: https://splunkbase.splunk.com/app/2890
+    version: 5.7.4
```
nasbench: Same as the above this is not needed since we removed the MLTK content.

install.yml
```diff
@@ -1,3 +1,48 @@
...
+    version: 3.7.2
+  - appid: Splunk_SA_Scientific_Python_linux_x86_64
+    url: https://splunkbase.splunk.com/app/2882
+    version: 4.3.2
+  - appid: SA-VMWIndex
+    url: https://splunkbase.splunk.com/app/5640
+    version: 4.0.4
+  - appid: Splunk_SA_CIM
+    url: https://splunkbase.splunk.com/app/1621
+    version: 8.5.0
```
nasbench:
```suggestion
    version: 8.5.0

```

nasbench (changes requested):
Overall LGTM except for the ML stuff. I do not know if we should host them in this instlall.yml

Also are we planning to add an automation that upgrades the versions automatically to the latest?

pyth0n1c:
> Also are we planning to add an automation that upgrades the versions automatically to the latest?
Yes - this work is almost done, but will land in a separate PR which:
1. bumps the contentctl-ng version, since this requires a contentctl-ng update
2. updates the workflows in this repo that handle that

nasbench (approved):

PR #4144, opened 2026-07-01
Swap order of macros in Attacker Tools On Endpoint

AndreiBanaru:
I was updating this morning my detections library and spotted this inconsistency.
The `drop_dm_object_name(1)` is always called before the `security_content_ctime(1)` macros.


patel-bhavin:
@AndreiBanaru - Thanks for fixing this inconsistency however it shouldn't matter functionally.

patel-bhavin (approved):

patel-bhavin:
Tested this change locally! 
image

patel-bhavin (approved):

PR #4145, opened 2026-07-01
Remote Thread Analytics Update

nasbench:
This PR revisits some the analytics that leverage Sysmon EID 8. In order to update following:

- Update the output fields of the data source to reflect meaningful output that is aligned with what the event report.
- Use the raw field names, due to a miss-attribution done by the Sysmon TA (for example, calling source/target processes as parent/child). A bug has been filed internally to address these issues.
- Deprecation of content that is super prone to FPs as well as merge content where it make sense to ease management.

### New Analytics [1]

- detections/endpoint/windows_uncommon_remote_thread_creation_in_browser_process.yml

### Updated Analytics [5]

- detections/endpoint/create_remote_thread_in_shell_application.yml
- detections/endpoint/create_remote_thread_into_lsass.yml
- detections/endpoint/powershell_remote_thread_to_known_windows_process.yml
- detections/endpoint/rundll32_create_remote_thread_to_a_process.yml
- detections/endpoint/windows_process_injection_remote_thread.yml

### Updated Data Source [1]

- data_sources/sysmon_eventid_8.yml

### Deprecated Analytics [3]

- detections/deprecated/rundll32_createremotethread_in_browser.yml
- detections/deprecated/windows_process_injection_of_wermgr_to_known_browser.yml
- detections/deprecated/windows_process_injection_with_public_source_path.yml

### DIFF
--- a/data_sources/sysmon_eventid_8.yml
+++ b/data_sources/sysmon_eventid_8.yml
@@ -1,8 +1,8 @@
 name: Sysmon EventID 8
 id: df7a786c-ade0-48f0-8596-26f10d169f7d
-version: 4
+version: 5
 creation_date: '2024-05-22'
-modification_date: '2026-05-13'
+modification_date: '2026-06-29'
 author: Patrick Bareiss, Splunk
 description: Logs the creation of a new thread in a process, including details about the thread ID, start address, and source process.
 mitre_components:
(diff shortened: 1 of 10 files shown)


patel-bhavin (approved):

patel-bhavin:
LGTM, generally speaking the TA should be fixed vs the searches. Now when the TA bug is resolved, we should update the searches to output CIM fields.

nasbench:
> LGTM, generally speaking the TA should be fixed vs the searches. Now when the TA bug is resolved, we should update the searches to output CIM fields.

I agree but keep in mind that the current limitation is a CIM one. AKA the TA cannot even map these correctly/accurately since CIM does not have the concept for these fields, but nonetheless, I will get them updated once things move forward.

PR #4146, opened 2026-07-05
Add PreAuthType filter to PetitPotam Kerberos TGT detection

### LINKED ISSUE #4141: [BUG]PetitPotam Suspicious Kerberos TGT Request — missing PreAuthType filter causes false positives on legitimate smart card logons
Summary
The detection PetitPotam Suspicious Kerberos TGT Request (Event 4768, MITRE T1187) filters on CertThumbprint!="" to flag certificate-based TGT requests, but does not filter on PreAuthType. As a result, the rule matches both PetitPotam exploitation attempts and legitimate smart card authentication, producing unnecessary noise for any environment where PKI/smart card logon is in normal use.

Current Logic
```
`wineventlog_security` EventCode=4768 src!="::1" TargetUserName=*$ CertThumbprint!=""
    BY dest, TargetUserName, src, action
| `petitpotam_suspicious_kerberos_tgt_request_filter`
```

Genuine smart card / certificate-based logons are expected to carry PreAuthType 15, 16, or 17 alongside a populated CertThumbprint. PetitPotam (via the ADCS/ESC8 relay path) abuses NTLM relay to force the DC to authenticate, ultimately yielding a TGT request where a certificate is present in the event but PreAuthType=2 — a combination that does not occur during normal smart-card use, since standard password pre-auth (type 2) doesn't carry certificate data.
This is confirmed by the project's own attack data sample for this technique:

[windows-xml-1.log](https://media.githubusercontent.com/media/splunk/attack_data/master/datasets/attack_techniques/T1187/petitpotam/windows-xml-1.log), where the captured event shows:

PreAuthType: 2
CertThumbprint: xyz

Because the current SPL only checks CertThumbprint!="", it cannot distinguish this anomalous combination from ordinary smart-card sign-ins (PreAuthType 15/16/17 + a real cert thumbprint), which are common in any enterprise using PKI-based authentication. This makes the detection noisy and reduces analyst trust in it.
Proposed Fix
Add an explicit PreAuthType=2 filter so the rule only fires on the anomalous combination (certificate field populated + password-style pre-auth type), which is the actual PetitPotam/ESC8 signature:

‍‍
wineventlog_security EventCode=4768 src!="::1" TargetUserName=*$ CertThumbprint!="" PreAuthType=2
    BY dest, TargetUserName, src, action
| `petitpotam_suspicious_kerberos_tgt_request_filter`

Impact

Before: Any 4768 event with a non-empty CertThumbprint matches, including legitimate smart card TGT requests (PreAuthType 15/16/17).
After: Only matches when a certificate field is present and the pre-auth type is 2 (password-style), which is the actual fingerprint of the PetitPotam coercion technique — reducing false positives without reducing true-positive coverage.

munzzyy:
### Details

Closes #4141.

The PetitPotam Suspicious Kerberos TGT Request analytic (`detections/endpoint/petitpotam_suspicious_kerberos_tgt_request.yml`) only filters on `CertThumbprint!=""`, which also matches ordinary smart card / PKI logons. Those show up as Event 4768 with a populated cert thumbprint too, so the rule fires on legitimate authentication in any environment that uses certificate-based logon.

The project's own attack data for this technique (`windows-xml-1.log` under `T1187/petitpotam`) has `PreAuthType` set to 2 alongside the cert thumbprint. That's the actual signature of the PetitPotam/ESC8 coercion path: password-style pre-auth (type 2) showing up with a certificate field populated. Normal smart card sign-in uses PreAuthType 15, 16, or 17 instead. I pulled the log myself and confirmed the field values rather than taking the issue's word for it.

Added `PreAuthType=2` to the search so it only matches on that specific combination.

### Testing

- Confirmed the current gap exists on `develop` HEAD by reading the detection file directly.
- Downloaded the referenced attack data log and verified `PreAuthType=2` and `CertThumbprint=xyz` are both present in the sample event, confirming the fix matches real data rather than a hypothetical.
- Ran `yamllint -c .yamllint` against the modified file. Passes clean.
- Ran a full `contentctl-ng build` against the repo with the change in place (2115 detections parsed, this one included). Built successfully, and the compiled `savedsearches.conf` shows the `PreAuthType=2` filter carried through correctly.
- Bumped `version` to 11 and `modification_date` to today, matching how other recent SPL-selection fixes in this repo have been versioned (e.g. #4112).

### Checklist

- [x] Validate name matches `<platform>_<mitre att&ck technique>_<short description>` nomenclature (unchanged, name/id preserved)
- [ ] CI/CD jobs passed (pending, see local build results above)
- [x] Validated SPL logic
- [x] Validated tags, description, and how to implement
- [x] Verified references match analytic
- [x] Confirm updates to lookups are handled properly (no lookup changes in this PR)

### DIFF
--- a/detections/endpoint/petitpotam_suspicious_kerberos_tgt_request.yml
+++ b/detections/endpoint/petitpotam_suspicious_kerberos_tgt_request.yml
@@ -1,8 +1,8 @@
 name: PetitPotam Suspicious Kerberos TGT Request
 id: e3ef244e-0a67-11ec-abf2-acde48001122
-version: 10
+version: 11
 creation_date: '2021-09-01'
-modification_date: '2026-05-13'
+modification_date: '2026-07-04'
 author: Michael Haag, Mauricio Velazco, Splunk
 status: production
 type: TTP
@@ -10,7 +10,7 @@ description: The following analytic detects a suspicious Kerberos Ticket Grantin
 data_source:
     - Windows Event Log Security 4768
 search: |-
-    `wineventlog_security` EventCode=4768 src!="::1" TargetUserName=*$ CertThumbprint!=""
+    `wineventlog_security` EventCode=4768 src!="::1" TargetUserName=*$ CertThumbprint!="" PreAuthType=2
       | stats count min(_time) as firstTime max(_time) as lastTime
         BY dest, TargetUserName, src,
            action
@@ -18,7 +18,7 @@ search: |-
       | `security_content_ctime(lastTime)`
       | `petitpotam_suspicious_kerberos_tgt_request_filter`
 how_to_implement: The following analytic requires Event Code 4768. Ensure that it is logging no Domain Controllers and appearing in Splunk.
-known_false_positives: False positives are possible if the environment is using certificates for authentication.
+known_false_positives: False positives are possible if the environment is using certificates for authentication, since PreAuthType 2 can appear on some initial or fallback authentication flows alongside a populated certificate field. Tune further based on observed authentication patterns in the environment.
 references:
     - https://www.ultimatewindowssecurity.com/securitylog/encyclopedia/event.aspx?eventid=4768
     - https://isc.sans.edu/forums/diary/Active+Directory+Certificate+Services+ADCS+PKI+domain+admin+vulnerability/27668/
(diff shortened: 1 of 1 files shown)


patel-bhavin:
@munzzyy - Thank you for proposing the fix. We are working on getting unit-testing to work on forks and will merge this PR after testing is done successfully.

nasbench (approved):
Verified the results in a lab.

Image

In addition https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/auditing/event-4768#kerberos-preauthentication-types showcases that all other types are not of interest to this rule.

Also the FP section showcases that the original rule authors simply forgot to add the conditions

PR #4147, opened 2026-07-05
Fix TargetUserName computer account filter in computer service ticket detection

### LINKED ISSUE #4083: [BUG] UserPrincipalName issue in Windows Large Number of Computer Service Tickets Requested
in the original search, filter must exclude the computer accounts but in action sometimes the output of 
TargetUserName is userPrincipalName like this scheme <sAMAccountName>$@<Domain Name>.<Domain TLD>
so the filter TargetUserName!="*$" is not effective.
for search effectivness add the following queries to the scenario to exclude computer accounts:

```
| eval TargetUserName = if(match(TargetUserName, "@"), mvindex(split(TargetUserName, "@"), 0), TargetUserName)
| search NOT TargetUserName = "*$"
```

this query will fix the computer account exclusion.

munzzyy:
### Details

Fixes #4083. The Windows Large Number of Computer Service Tickets Requested detection excludes computer accounts with `TargetUserName!="*$"`, but Event 4769 logs `TargetUserName` in userPrincipalName form (`sAMAccountName@domain.tld`) in some environments instead of the plain sAMAccountName. When that happens the trailing `$` is gone and the filter never fires, so computer accounts still count toward the "large number of tickets" total and inflate the detection with false positives.

Fix strips everything after `@` before checking for the `$` suffix, same pattern already used in `suspicious_kerberos_service_ticket_request.yml` for the same field on the same event:

```
| eval TargetUserName = mvindex(split(TargetUserName, "@"), 0)
| search NOT TargetUserName="*$"
```

Bumped version to 12 and updated modification_date.

### Checklist

- [x] Validated SPL logic.
- [x] Verified references match analytic.
- [x] Confirm updates to lookups are handled properly.


patel-bhavin:
@munzzyy - Thank you for proposing the fix. We are working on getting unit-testing to work on forks and will merge this PR after testing is done successfully.

munzzyy:
Sounds good, thanks for the heads up. No rush on my end - let me know if there's anything you'd like me to change in the meantime.

nasbench (approved):
New update is working correctly. Here a screenshot

Image

I also added the condition `TargetUserName!="*$"` back, to filter early and enhance perf

PR #4151, opened 2026-07-06
Only run unit-testing when there's something to test

ljstella:
### Details

Tweak to the `unit-testing` workflow so that we only occupy the larger runner that testing uses when we actually have things to test, as opposed to every PR.

This change is needed because the multi-step testing process w/ -ng now builds the test environment first, and then when running testing determines what needs to be tested, as opposed to determining if testing even needs to happen first.


pyth0n1c:
I talked with @ljstella about this. Eventually we should handle this type of logic in the tool test workflow, but this is a good catch-all for now to ensure we don't waste time setting up environments when we won't actually have any content to test.

I will defer to STRT for approval, but concur with the change.

patel-bhavin (approved):

.github/workflows/unit-testing.yml
```diff
@@ -2,6 +2,11 @@ name: unit-testing
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+    paths:
+      - detections/**
+      - data_sources/**
+      - lookups/**
+      - macros/**
```
nasbench, line 9: What about stories?

patel-bhavin: i think changes to the story files shouldn't trigger testing, unless you are changing the name, in which case, there will be  changes to the detection yamls too.

ljstella: Yeah, changing a story's description or adding an additional reference doesn't cause the associated detections to change or trigger testing.

nasbench (approved):

.github/workflows/unit-testing.yml
nasbench: My point was if you are changing a story title/rename it, or break it in whatever way, no tests will be triggered. I know this is rare case but this will break things without us knowing

ljstella: That doesn't trigger unit testing anyway though, but would cause the build step to fail, which is still run via a separate job.

nasbench: Nevermind, I just saw that this unit-testing not build (facepalm).

PR #4161, opened 2026-07-09
Bug Fix - Suspicious Process Executed From Container File

patel-bhavin:
This PR fixes a field-mapping issue in the suspicious process executed from container file detection.

The detection previously extracted the file launched from inside the archive into process_name, which could overwrite or conflict with the CIM process_name field. Since process_name should represent the actual process image, such as wscript.exe, this update renames the extracted archive member field to container_file_name.

This keeps CIM process fields accurate while still capturing the suspicious file executed from the compressed container.

Updated behavior:
```
process_name = actual executing process
file_name = compressed container file
container_file_name = file launched from inside the container
```

Slack - https://splunk.slack.com/archives/C9RE1GUSZ/p1783544914063179

### DIFF
--- a/detections/endpoint/suspicious_process_executed_from_container_file.yml
+++ b/detections/endpoint/suspicious_process_executed_from_container_file.yml
@@ -19,21 +19,13 @@ search: |-
     from datamodel=Endpoint.Processes
     where Processes.process IN (
         "*.ZIP\\*",
-        "*.zip\\*",
         "*.ISO\\*",
-        "*.iso\\*",
         "*.IMG\\*",
-        "*.img\\*",
         "*.CAB\\*",
-        "*.cab\\*",
         "*.TAR\\*",
-        "*.tar\\*",
         "*.GZ\\*",
-        "*.gz\\*",
         "*.RAR\\*",
-        "*.rar\\*",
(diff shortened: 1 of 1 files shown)


nasbench (dismissed):

detections/endpoint/suspicious_process_executed_from_container_file.yml
```diff
@@ -11,7 +11,58 @@ data_source:
...
+        "*.CAB\\*",
+        "*.cab\\*",
+        "*.TAR\\*",
+        "*.tar\\*",
+        "*.GZ\\*",
+        "*.gz\\*",
+        "*.RAR\\*",
+        "*.rar\\*",
+        "*.7Z\\*",
+        "*.7z\\*"
```
nasbench: Why include both casing? Its already accounted for since the compare is case insensitive. Please revert your last commit

patel-bhavin: done! had added one of them for testing something - seems like during formatting the AI tool ended up lowercase all of them :\. Should be fixed now!

nasbench (approved):

nasbench (approved):

PR #4184, opened 2026-07-27
removing dashbaord due to VULN

patel-bhavin:
Removing this dashboard entirely due to the reported VULN (deadline 7/31) and the searches in the dashboard do not work since the TA needs fixing!


pyth0n1c (approved):
Agreed, this should simply be removed for a number of reasons discussed internally.

ljstella (approved):
LGTM

PR #4189, opened 2026-07-28
Minor fixes

nasbench:
Minor fixes


pyth0n1c (approved):
Since this is a description update and a brief additive search update, I am comfortable approving this even though unit testing has not finished yet.

PR #4191, opened 2026-07-29
Bump contentctl.yml and build.yml to 6.4.0

patel-bhavin:
Automated bump after v6.3.0.


nasbench (dismissed):
LGTM, need only changing the deprecated to removed.

nasbench (approved):

PR #4198, opened 2026-08-04
Remove JS from ESCU

P4T12ICK:


tccontre (dismissed):
LGTM

nasbench (approved):
LGTM! After double checking with Patrick, we confirmed its working.

PR #4202, opened 2026-08-05
Fix Issues 6.4 Batch

### LINKED ISSUE #4188: [BUG] Logic Problem in "Windows Disable Shutdown Button Through Registry"
The current detection alerts when the shutdownwithoutlogon registry key is set to 0x00000000. However, setting this to 0 is the default Windows Server behavior and a strict CIS Benchmark requirement:
(2.3.13.1 (L1) Ensure 'Shutdown: Allow system to be shut down without having to log on' is set to 'Disabled' (Automated)) 
to prevent unauthenticated lock-screen shutdowns.

The Trend Micro malware [reference](https://www.trendmicro.com/vinfo/us/threat-encyclopedia/malware/ransom.msil.screenlocker.a/) in the description was likely misinterpreted. The article lists setting the value to 0 as the remediation/restore step to fix the machine, not the malicious state.

I suggest changing the shutdownwithoutlogon condition to monitor for the 0x00000001 value. This will correctly catch the malware behavior noted by Trend Micro, while also alerting on servers that are violating CIS best practices.

Thanks!

### LINKED ISSUE #4199: [BUG] typo in "Windows Registry Dotnet ETW Disabled Via ENV Variable"
Hey team,

There’s a simple typo in this detection. The `Registry.registry_value_data` value has 9 digits (`0x000000000`), and it isn’t enclosed in double quotes. :D 

Thanks!

### LINKED ISSUE #4201: [BUG]Executable File Written in Administrative SMB Share" uses exact-match AccessMask, missing combined bitmask values
## Search affected
`Executable File Written in Administrative SMB Share`
(EventCode=5145, admin share write detection)

## Current logic
```spl
`wineventlog_security` EventCode=5145 RelativeTargetName IN ("*.exe","*.dll")
ObjectType=File ShareName IN ("\\\\*\\C$","\\\\*\\IPC$","\\\\*\\admin$")
AccessMask="0x2"
```

## Problem
`AccessMask` in EventCode 5145 is a **bitmask**, not an enum — Windows OR's together every access right requested in a single hex value (per the standard file/pipe access-right bit layout: `0x1`=ReadData, `0x2`=WriteData, `0x4`=AppendData, `0x20`=Execute, `0x10000`=Delete, `0x20000`=ReadControl, `0x100000`=Synchronize, etc.).

In practice, SMB clients (including the tooling attackers actually use — `psexec`, `smbclient`, Impacket's `smbclient.py`, CrackMapExec, native `net use` + `copy`) rarely request WriteData in isolation. The logged `AccessMask` is commonly something like:

- `0x100002` (WriteData + Synchronize)
- `0x12019f` (generic write/full access bundle)
- `0x1301bf`

None of these string-match `"0x2"`, so the rule silently fails to fire on the majority of real-world write events. This is a straightforward, low-effort detection bypass — an attacker doesn't even need to evade deliberately; default tooling behavior already avoids the exact-match condition.

## Reproduction
1. From an attacker host, drop an `.exe`/`.dll` onto a target's `C$`/`ADMIN$` share using any standard SMB client (e.g. `smbclient.py target -c "put implant.exe"` or `copy implant.exe \\target\C$\Windows\Temp\`).
2. Inspect the resulting EventCode 5145 on the target — `AccessMask` will almost always be a combined value (e.g. `0x100002`), not `0x2`.
3. Confirm the ESCU search does not fire despite a genuine executable write to the admin share.

## Suggested fix
Replace the exact string match with a bitwise check on the WriteData bit (and optionally AppendData, since some write paths use `0x4`). SPL supports bitwise operators in `eval`/`where`:

```spl
`wineventlog_security` EventCode=5145 RelativeTargetName IN ("*.exe","*.dll")
ObjectType=File ShareName IN ("\\\\*\\C$","\\\\*\\IPC$","\\\\*\\admin$")
| eval AccessMask_dec = tonumber(AccessMask, 16)
| where (bit_and(AccessMask_dec, 2) > 0) OR (bit_and(AccessMask_dec, 4) > 0)
| stats min(_time) as firstTime max(_time) as lastTime count by EventCode ShareName RelativeTargetName ObjectType AccessMask src_user src_port IpAddress dest
| `executable_file_written_in_administrative_smb_share_filter`
```

This preserves intent (flag any write to an exe/dll on an admin share) while correctly matching AccessMask values where the WriteData (`0x2`) or AppendData (`0x4`) bit is set, regardless of what else is OR'd in.

## Impact
- **Severity:** High for this detection's stated purpose — it's a full logic bypass, not an edge case. Most legitimate SMB write clients set additional bits by default, so the false-negative rate in practice is likely very high.
- **Detection coverage lost:** Lateral movement / remote implant staging via admin shares (a common precursor to PsExec-style execution) goes undetected.

nasbench:
This PR fixes a couple of issues reported by the community through GH and E-mail, as well as a bunch of other issues identified by STRT.


tccontre (dismissed):
LGTM

detections/endpoint/windows_user_discovery_via_net.yml
```diff
@@ -1,61 +1,87 @@
...
+        "*user",
+        "*users *",
+        "*users"
+    )
+    NOT Processes.process IN (
+        "*/add*",
+        "*/delete*",
+        "*/active*",
+        "*/times*",
+        "*/delete*",
```
nasbench, line 35:
```suggestion
        "*/add*",
        "*/active*",
        "*/delete*",
        "*/times*"
```

tccontre (approved):
looks good

PR #4203, opened 2026-08-05
pin windows ta version in updater

pyth0n1c:
pin to current version, intentionally clobbering any updates to the windows ta


nasbench (approved):
LGTM - CI will timeout due to the large amount of rules to be tested. Will confirm this on the internal side.

PR #4205, opened 2026-08-07
Detections against malicious Python package activities

onurmerdogan:
### Details

This PR contains detections against malicious activities during Python package installations.

### Stories
- **Malicious Python Package Installation:** New story in order to track malicious activities during Python package installations. Covers both installation and build related activities.

### Detections
- **Python Network Traffic during Package Build:** Detects a Python process making an outbound network connection during package installation.
- **Python PTH File Creation during Package Installation:** detects the creation of a Python path configuration (`.pth`) file in conjunction with a package installation process.
- **Python PYTHONPATH Modification During Package Installation:** Detects modification of the PYTHONPATH environment variable in conjunction with a package installation process.
- **Python Site Hooks Creation during Package Installation:** Detects the creation of a Python site hook file  within a packages directory in conjunction with a package installation process.


detections/endpoint/python_network_traffic_during_package_build.yml
```diff
@@ -0,0 +1,67 @@
...
+    because it allows adversaries to establish a foothold or exfiltrate data without any direct interaction from the victim beyond
+    running `pip install`. If confirmed malicious, this could indicate a successful software supply chain compromise.
+data_source:
+    - Sysmon EventID 1
+    - Sysmon EventID 3
+search: |-
+    `sysmon` EventID IN (1,3)
+    | bin _time span=5m
+    | stats values(parent_process_id) as parent_process_id, values(dest_ip) as dest_ip, values(QueryName) as QueryName, values(CommandLine) as CommandLine, values(parent_process) as parent_process, dc(EventID) as EventID count by _time, host, source, process_id
+    | search CommandLine="*_in_process.py*" AND CommandLine="* build_wheel*" AND dest_ip!=""
```
patel-bhavin: @onurmerdogan : it would be better to use these filters earlier in the search like :

```
`sysmon` EventID IN (1,3) process="*_in_process.py*" AND process="* build_wheel*" AND dest_ip!=""
.....
.....
```

patel-bhavin: host usually refers to the machine sending the data : i would consider using `dest` in the by clause

detections/endpoint/python_network_traffic_during_package_build.yml
```diff
@@ -0,0 +1,67 @@
...
+    | `python_network_traffic_during_package_build_filter`
+how_to_implement: |-
+    The detection is based on data that originates from Sysmon. To implement this search, you must ingest logs
+    with process creation (EventID 1) and network connection (EventID 3) events, mapped via the appropriate
+    Splunk Technology Add-on. Use the Splunk Common Information Model (CIM) to normalize the field names.
+known_false_positives: Python packages may contact software repositories, mirror sites during build time. Investigate the destination and package content to determine legitimacy.
+references:
+    - https://blog.talosintelligence.com/the-serpents-tongue-luring-the-python-out-of-its-den/
+drilldown_searches:
+    - name: View the detection results for - "$host$"
```
patel-bhavin: use `dest`

detections/endpoint/python_network_traffic_during_package_build.yml
```diff
@@ -0,0 +1,67 @@
+name: Python Network Traffic during Package Build
```
nasbench:
```suggestion
name: Python Network Traffic During Package Build
```

detections/endpoint/python_network_traffic_during_package_build.yml
```diff
@@ -0,0 +1,67 @@
...
+id: 03c9c504-2294-44da-8180-beefe1ca8ba8
+version: 1
+creation_date: '2026-07-23'
+modification_date: '2026-07-23'
+author: Onur Mustafa Erdogan, Splunk
+status: production
+type: TTP
```
nasbench: I am assuming you did check FP rates, but if not, please do. Since TTP will generate an alert every time. Perhaps this is worth an anomaly imo.

detections/endpoint/python_network_traffic_during_package_build.yml
```diff
@@ -0,0 +1,67 @@
...
+description: |-
+    The following analytic detects a Python process making an outbound network connection during package installation.
+    Adversaries can abuse `setup.py` build scripts by leveraging `distutils`/`setuptools` command classes to execute arbitrary code,
+    including network beacons to third-party domains, the moment a malicious Python package is installed. This activity is significant
+    because it allows adversaries to establish a foothold or exfiltrate data without any direct interaction from the victim beyond
+    running `pip install`. If confirmed malicious, this could indicate a successful software supply chain compromise.
+data_source:
+    - Sysmon EventID 1
+    - Sysmon EventID 3
```
nasbench: There is a convention where if the DS are both required we use an `AND`
```suggestion
    - Sysmon EventID 1 AND Sysmon EventID 3
```

detections/endpoint/python_network_traffic_during_package_build.yml
```diff
@@ -0,0 +1,67 @@
...
+search: |-
+    `sysmon` EventID IN (1,3)
+    | bin _time span=5m
+    | stats values(parent_process_id) as parent_process_id, values(dest_ip) as dest_ip, values(QueryName) as QueryName, values(CommandLine) as CommandLine, values(parent_process) as parent_process, dc(EventID) as EventID count by _time, host, source, process_id
```
nasbench: QueryName is not available in Sysmon EID 3, so perhaps you want to remove that.

detections/endpoint/python_network_traffic_during_package_build.yml
nasbench: Instead of using CommandLine (which is fine). Let's use the CIM names where possible since those are more friendly to ES and its feature.

So in this example we use `process` instead of CommandLine

detections/endpoint/python_network_traffic_during_package_build.yml
```diff
@@ -0,0 +1,67 @@
...
+    - name: View the detection results for - "$host$"
+      search: '%original_detection_search% | search  host = "$host$"'
+      earliest_offset: $info_min_time$
+      latest_offset: $info_max_time$
+    - name: View risk events for the last 7 days for - "$host$"
+      search: '| from datamodel Risk.All_Risk | search normalized_risk_object IN ("$host$") | stats count min(_time) as firstTime max(_time) as lastTime values(search_name) as "Search Name" values(risk_message) as "Risk Message" values(analyticstories) as "Analytic Stories" values(annotations._all) as "Annotations" values(annotations.mitre_attack.mitre_tactic) as "ATT&CK Tactics" by normalized_risk_object | `security_content_ctime(firstTime)` | `security_content_ctime(lastTime)`'
+      earliest_offset: 7d
+      latest_offset: "0"
+finding:
+    title: A Python process on $host$ made an outbound network connection to $dest$ during package installation
```
nasbench: Try and include more info in the message such as CommandLine/Parent Commandline ...etc. Since this will be what is shown to the user in ES MC

detections/endpoint/python_network_traffic_during_package_build.yml
```diff
@@ -0,0 +1,67 @@
...
+    entity:
+        field: host
+        type: system
+        score: 30
```
nasbench: Consider adding a threat_objects fiel with the `process` field as the object.

detections/endpoint/python_pth_file_creation_during_package_installation.yml
```diff
@@ -0,0 +1,70 @@
+name: Python PTH File Creation during Package Installation
```
nasbench:
```suggestion
name: Python PTH File Creation During Package Installation
```

detections/endpoint/python_network_traffic_during_package_build.yml
nasbench: Also to optimize the search further, I think it make sense to add filter for the Image? Without it, you will be grabbing a lot of EID 3 events with only `dest_ip != ""`

detections/endpoint/python_pth_file_creation_during_package_installation.yml
```diff
@@ -0,0 +1,70 @@
...
+id: bf581b86-39cd-48b7-9312-d9affb48a8bc
+version: 1
+creation_date: '2026-07-23'
+modification_date: '2026-07-23'
+author: Onur Mustafa Erdogan, Splunk
+status: production
+type: TTP
```
nasbench: Same comment here. Do you want this to be an alert or more like something suspicious  that increases risk?

detections/endpoint/python_pth_file_creation_during_package_installation.yml
```diff
@@ -0,0 +1,70 @@
...
+    with a package installation process. Path configuration files placed under `site-packages` or `dist-packages`
+    are executed with every subsequent invocation of Python, allowing adversaries to achieve persistence
+    on the victim endpoint regardless of build method or distribution type. This technique was used by the
+    threat actor group TeamPCP during the supply chain compromise of the `litellm` package. If confirmed
+    malicious, this could result in arbitrary code execution every time Python is invoked on the compromised host.
+data_source:
+    - Sysmon EventID 1
+    - Sysmon EventID 11
+search: |-
+    `sysmon` EventID IN (1,11) AND (CommandLine="* install *" OR (TargetFilename="*.pth" AND action="created"))
```
nasbench: Same comment as above, let's try and use the CIM field where possible.
Also does it make sense to add an process_name filter to enhance the performance of the search?

detections/endpoint/python_pth_file_creation_during_package_installation.yml
```diff
@@ -0,0 +1,70 @@
...
+    - name: View risk events for the last 7 days for - "$host$"
+      search: '| from datamodel Risk.All_Risk | search normalized_risk_object IN ("$host$") | stats count min(_time) as firstTime max(_time) as lastTime values(search_name) as "Search Name" values(risk_message) as "Risk Message" values(analyticstories) as "Analytic Stories" values(annotations._all) as "Annotations" values(annotations.mitre_attack.mitre_tactic) as "ATT&CK Tactics" by normalized_risk_object | `security_content_ctime(firstTime)` | `security_content_ctime(lastTime)`'
+      earliest_offset: 7d
+      latest_offset: "0"
+finding:
+    title: A Python `.pth` file called $TargetFilename$ was created on $host$ during package installation
+    entity:
+        field: host
+        type: system
+        score: 40
```
nasbench: Same comment as above for the message and missing threat objects

detections/endpoint/python_pythonpath_modification_during_package_installation.yml
```diff
@@ -0,0 +1,74 @@
+name: Python PYTHONPATH Modification During Package Installation
+id: 00f01ba4-df01-4caa-90f6-187e4563a516
+version: 1
+creation_date: '2026-07-23'
+modification_date: '2026-07-23'
+author: Onur Mustafa Erdogan, Splunk
+status: production
+type: TTP
```
nasbench, line 8: Same comment as above for the type

detections/endpoint/python_pythonpath_modification_during_package_installation.yml
```diff
@@ -0,0 +1,74 @@
...
+    sessions. If confirmed malicious, this could result in arbitrary code execution every time Python
+    is invoked by the affected user.
+data_source:
+    - Sysmon EventID 1
+    - Sysmon EventID 13
+search: |-
+    `sysmon` EventID IN (1,13) AND (CommandLine="* install *" OR (TargetObject="*PYTHONPATH" AND action="modified"))
+    | bin _time span=5m
+    | stats values(parent_process_id) as parent_process_id, values(TargetObject) as TargetObject, values(registry_value_data) as registry_value_data, values(CommandLine) as CommandLine, values(parent_process) as parent_process, dc(EventID) as dc_event_id count by _time, host, source, process_id
+    | search dc_event_id>1
```
nasbench: Same comment for the field names and filters

nasbench (changes requested):

detections/endpoint/python_pythonpath_modification_during_package_installation.yml
onurmerdogan: This one is quite rare activity, given registry key is not often not monitored. Also, editing this key on Windows seem to be quite rare, I suggest still keeping it as TTP, but if you like we can get this one downgraded to anomaly, what do you think @patel-bhavin  @nasbench

nasbench: We can keep it to TTP since we have data on it.

nasbench (approved):
LGTM.

I made a couple of changes to the output format via `table` to make the results consistent.

And in the `Python Network Traffic During Package Build` since there is a bug in EID 3 mapping of `dest` (it maps it to the actual dest from the event instead of the host). I made some changes to make it usable in the `by` clause.

PR #4209, opened 2026-08-10
The following attack data links were partially bad, but not totally wrong.

pyth0n1c:
Several of them had a path containing double backslash, like attack_data//T1003/file.log. This is resolvable by a redirect, but still wrong. A few others did not point to the prefix https://media.githubusercontent.com/media/splunk/attack_data/master/datasets and instead started with https://media.githubusercontent.com/media/splunk/attack_data/refs/head/master/datasets which, while it still resolved, was wrong

**DETECTION VERSIONS ARE INTENTIONALLY NOT BUMPED BECAUSE THEY SHOULDN'T BE - ATTACK DATA LINKS ARE NOT SERIALIZED INTO SAVEDSEARCHES.CONF**


nasbench (approved):

nasbench (commented):
Assuming this is targeting 6.5. So I will not merge for now

pyth0n1c:
> Assuming this is targeting 6.5. So I will not merge for now

Correct. This also doesn't actually matter for releases, it is only relevant to validation/testing workflows which validate attack data links. Interesting enough, you will find both links to an attack_data file "work":
https://media.githubusercontent.com/media/splunk/attack_data/master/datasets/attack_techniques/T1546.012/atomic_red_team/windows-application.log

https://media.githubusercontent.com/media/////splunk/attack_data///////master/datasets/attack_techniques//////T1546.012/atomic_red_team/windows-application.log

It's just that the second is "silently" redirected.


And I suppose different sites handle this differently. Both 
https://www.google.com/maps
and 
https://www.google.com/////maps
image

work, with the second one NOT being redirected.

Interesting...

PR #4212, opened 2026-08-11
Fix Issues & Deprecate Some Experimental Content

### LINKED ISSUE #4101: [BUG]Poor Logic of Windows EDRSilencer Execution
Hello Splunk Security Team,
based on my research on the EDRSicencer tool and its functionality in its github repo located at https://github.com/netero1010/EDRSilencer/blob/main/EDRSilencer.c#L128
the scenario query is based on EDRSilencer.exe process name that is not efficient and in repo compile section recommend this name and if i compile the binary as follow the rule will completely bypass:
x86_64-w64-mingw32-gcc EDRSilencer.c utils.c -o <random name>.exe -lfwpuclnt
i have fully tested the compiled binary on windows that will add WPF filter on msmpeng.exe process
the rule can be converted to the add new WPF filter EID.
i have sample logs of this activity that use can write more efficient scenario based on this:

this is windows log
```
[-](https://github.com/splunk/security_content/issues#) <Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
[-](https://github.com/splunk/security_content/issues#) <System>
  <Provider Name="Microsoft-Windows-Security-Auditing" Guid="{54849625-5478-4994-a5ba-3e3b0328c30d}" /> 
  <EventID>5447</EventID> 
  <Version>0</Version> 
  <Level>0</Level> 
  <Task>13573</Task> 
  <Opcode>0</Opcode> 
  <Keywords>0x8020000000000000</Keywords> 
  <TimeCreated SystemTime="2026-05-23T15:32:49.8857017Z" /> 
  <EventRecordID>153502136</EventRecordID> 
  <Correlation ActivityID="{a015d80e-e931-0006-0cd9-15a031e9dc01}" /> 
  <Execution ProcessID="1152" ThreadID="900" /> 
  <Channel>Security</Channel> 
  <Computer>SOC[-](https://github.com/splunk/security_content/issues#)20.soc.local</Computer> 
  <Security /> 
  </System>
- <EventData>
  <Data Name="ProcessId">16824</Data> 
  <Data Name="UserSid">S-1-5-21-3258699341-4129412386-1304611611-1618</Data> 
  <Data Name="UserName">SOC\m.hamedani</Data> 
  <Data Name="ProviderKey">{a4b43ab7-75d5-4d88-a62c-8b5d2677c73d}</Data> 
  <Data Name="ProviderName">Microsoft Corporation</Data> 
  <Data Name="ChangeType">%%16384</Data> 
  <Data Name="FilterKey">{da607736-f157-4bdd-9ad9-042516072b55}</Data> 
  <Data Name="FilterName">Custom Outbound Filter</Data> 
  <Data Name="FilterType">%%16387</Data> 
  <Data Name="FilterId">305460</Data> 
  <Data Name="LayerKey">{c38d57d1-05a7-4c33-904f-7fbceee60e82}</Data> 
  <Data Name="LayerName">ALE Connect v4 Layer</Data> 
  <Data Name="LayerId">48</Data> 
  <Data Name="Weight">18446744073709551615</Data> 
  <Data Name="Conditions">Condition ID: {d78e1e87-8644-4ea5-9437-d809ecefc971} Match value: Equal to Condition value: 00000000 5c 00 64 00 65 00 76 00-69 00 63 00 65 00 5c 00 \.d.e.v.i.c.e.\. 00000010 68 00 61 00 72 00 64 00-64 00 69 00 73 00 6b 00 h.a.r.d.d.i.s.k. 00000020 76 00 6f 00 6c 00 75 00-6d 00 65 00 33 00 5c 00 v.o.l.u.m.e.3.\. 00000030 70 00 72 00 6f 00 67 00-72 00 61 00 6d 00 64 00 p.r.o.g.r.a.m.d. 00000040 61 00 74 00 61 00 5c 00-6d 00 69 00 63 00 72 00 a.t.a.\.m.i.c.r. 00000050 6f 00 73 00 6f 00 66 00-74 00 5c 00 77 00 69 00 o.s.o.f.t.\.w.i. 00000060 6e 00 64 00 6f 00 77 00-73 00 20 00 64 00 65 00 n.d.o.w.s. .d.e. 00000070 66 00 65 00 6e 00 64 00-65 00 72 00 5c 00 70 00 f.e.n.d.e.r.\.p. 00000080 6c 00 61 00 74 00 66 00-6f 00 72 00 6d 00 5c 00 l.a.t.f.o.r.m.\. 00000090 34 00 2e 00 31 00 38 00-2e 00 32 00 36 00 30 00 4...1.8...2.6.0. 000000a0 34 00 30 00 2e 00 37 00-2d 00 30 00 5c 00 6d 00 4.0...7.-.0.\.m. 000000b0 73 00 6d 00 70 00 65 00-6e 00 67 00 2e 00 65 00 s.m.p.e.n.g...e. 000000c0 78 00 65 00 00 00 x.e...</Data> 
  <Data Name="Action">%%16389</Data> 
  <Data Name="CalloutKey">{00000000-0000-0000-0000-000000000000}</Data> 
  <Data Name="CalloutName">-</Data> 
  </EventData>
  </Event>
```

and on sysmon with this path that will show us persistent WPF policy
HKLM\System\CurrentControlSet\Services\BFE\Parameters\Policy\Persistent\Filter\

### LINKED ISSUE #4210: [BUG] "Windows Set Account Password Policy To Unlimited Via Net" - Missing wildcard
Hi Splunk team,

I suspect that a wildcard is missing in the second process filter for 
[Windows Set Account Password Policy To Unlimited Via Net](https://research.splunk.com/endpoint/11f93009-8083-43fd-82a7-821fcbdc8342/)

(Processes.process="* /maxpwage:unlimited" OR Processes.process="**\*** /maxpwage:49710")

### LINKED ISSUE #4211: Suggested Enhancement for Detection: Ping Sleep Batch Command
The current detection is using `Processes.parent_process = "*-n*"`, but does not detect the usage of `Processes.parent_process = "*/n*"`

Detection enhancement by filtering for both slash and hyphen notation to detect alternative spellings. Or is there a special reason why hyphen notation is not taken into account?

nasbench:
TBD


detections/endpoint/ping_sleep_batch_command.yml
onurmerdogan, line 15: Should we consider adding `Windows Event Log Security 4688` here also?

detections/endpoint/ping_sleep_batch_command.yml
```diff
@@ -22,7 +22,10 @@ search: |-
     FROM datamodel=Endpoint.Processes WHERE
     (
         Processes.parent_process= "*ping*"
-        Processes.parent_process = *-n*
+        Processes.parent_process IN (
+            "*-n*",
+            "*/n*"
+        )
         Processes.parent_process IN (
```
onurmerdogan, line 30: In the past, we've observed use of double pipe, `||`, it might be good to incorporate.

detections/endpoint/ping_sleep_batch_command.yml
```diff
@@ -38,7 +41,10 @@ search: |-
             OR
             Processes.original_file_name= "ping.exe"
         )
-        Processes.process = *-n*
+        Processes.process IN (
+            "*-n*",
+            "*/n*"
+        )
         Processes.process IN (
```
onurmerdogan, line 50: Same here, `||`, it might be good to incorporate.

detections/endpoint/ping_sleep_batch_command.yml
onurmerdogan: Little side note here, chained commands on command interpreters often gets logged individually, rather than getting logged as they were typed on the interpreter. For instance,
`ping -n 1.2.3.4.5 || hostname` would be logged in two separate logs, one for `ping.exe` and one for `hostname.exe`.

I've seen EDR's capturing issued commands as part of process activity logs also, in those cases this detection would prove the most useful I believe. It might be good to mention on implementation part maybe?

detections/endpoint/windows_set_account_password_policy_to_unlimited_via_net.yml
```diff
@@ -1,38 +1,49 @@
...
-         Processes.user_id Processes.vendor_product
+    | tstats `security_content_summariesonly`
+      values(Processes.process) as process min(_time) as firstTime max(_time) as lastTime
+
+    FROM datamodel=Endpoint.Processes WHERE
+
+    `process_net`
+    Processes.process="* accounts *"
+    Processes.process IN (
+        "* /maxpwage:49710*",
```
onurmerdogan: `maxpwage` range is reduced to range of 1-999.
https://learn.microsoft.com/en-us/troubleshoot/windows-server/networking/net-commands-on-operating-systems

Does it make sense to apply some regex instead? We can filter initially for the existence of `/maxpwage`, and after BY clauses, we can add regex filter for `\/maxpwage\:(?:[3-9]\d{2}|unlimited)`.

detections/endpoint/windows_set_account_password_policy_to_unlimited_via_net.yml
onurmerdogan, line 68: Do we need any threat_objects here?

detections/endpoint/ping_sleep_batch_command.yml
onurmerdogan, line 109: Do we need any threat_objects here?

onurmerdogan (commented):
Looks great, just a few minor comments and questions I have.

detections/endpoint/windows_set_account_password_policy_to_unlimited_via_net.yml
nasbench: PR is in draft, not all additions are included but thanks for checking. TO will be included

detections/endpoint/ping_sleep_batch_command.yml
nasbench: PR is in draft, not all additions are included but thanks for checking. TO will be included

nasbench: Yes I am aware, the detection has the intent to capture the thing when its executed via flags such as /c or -command, etc. Nothing to be controlled by the user since thats how windows handles things.

detections/endpoint/ping_sleep_batch_command.yml
nasbench: Yes technically its a good thing to add, but since the double pipe is used to check for failure of the left hand side to run the right hand side. Since the logic is going with the idea of doing a ping as a sleep mechanism, it does not make sense to fail the left command.

I will add it however just because its an easy add

detections/endpoint/ping_sleep_batch_command.yml
nasbench: See my response above

detections/endpoint/windows_set_account_password_policy_to_unlimited_via_net.yml
nasbench: Thanks for checking this out. Yeah the regex approach will work too. I will incorporate that.

detections/endpoint/ping_sleep_batch_command.yml
nasbench: No, since that data source does not contain a parent command line entry

onurmerdogan (approved):
Overall looks good to me. There are few thins we can improve on future PRs, (Event ID 5441 fields and example log, more config file for Linux detection etc.) nothing urgent.

P4T12ICK (approved):
LGTM

PR #4223, opened 2026-08-17
Environments for internal/external unit testing

pyth0n1c:
run unit testing in a a different environment, manually approved by security_content repo member after reviewing PR to enable full testing with appropriate Apps.

TBD, still in progress


nasbench (commented):
While you are it @pyth0n1c please do the same for https://github.com/splunk/attack_data since we also removed token access for ext contribs there too.

PR #4224, opened 2026-08-17
Bump to contentctl 1.1.0 to facilitate testing

pyth0n1c:
DO NOT MERGE - this is a bump created for testing purposes only.


nasbench:
This is no longer needed since NG was bumped here - https://github.com/splunk/security_content/pull/4232

PR #4245, opened 2026-08-30
Fix original_file_name field typo in whoami and arp discovery detections

munzzyy:
### Details

Both searches OR two conditions to catch a renamed LOLBin: `process_name`, or the PE's original file name field (survives a rename). The second condition references `Processes.process_original_file_name`. That field does not exist in the CIM Endpoint.Processes model. The correct field is `Processes.original_file_name`, already used correctly a few lines down in each detection's own BY clause.

A field with no indexed values contributes nothing to a tstats OR. Both detections quietly collapse to a plain `process_name="whoami.exe"` or `process_name="arp.exe"` check. Rename the binary and both miss it. Catching that exact case is the whole point of checking original_file_name.

Fixed files:
- `detections/endpoint/system_user_discovery_with_whoami.yml`
- `detections/endpoint/network_connection_discovery_with_arp.yml`

Checked against the CIM 5.3 Endpoint docs, `original_file_name` is the only valid field. Bumped version and modification_date on both.

Ran `contentctl-ng build` locally against develop with these files changed. All 2157 detections parsed clean, corrected SPL renders right in savedsearches.conf. No Splunk instance here to replay the attack_data tests, so the True Positive test is unconfirmed end to end. Happy to add anything else the checklist needs, thanks.

### Checklist

- [ ] Validate name matches `<platform>_<mitre att&ck technique>_<short description>` nomenclature - N/A, no detection added or renamed
- [ ] [CI/CD](https://github.com/splunk/security_content/actions) jobs passed - ran `contentctl-ng build` locally instead, see note above
- [x] Validated SPL logic.
- [ ] Validated tags, description, and how to implement. - N/A, unchanged
- [ ] Verified references match analytic. - N/A, unchanged
- [ ] Confirm updates to lookups are handled properly. - N/A, no lookups touched

### DIFF
--- a/detections/endpoint/network_connection_discovery_with_arp.yml
+++ b/detections/endpoint/network_connection_discovery_with_arp.yml
@@ -1,8 +1,8 @@
 name: Network Connection Discovery With Arp
 id: ae008c0f-83bd-4ed4-9350-98d4328e15d2
-version: 10
+version: 11
 creation_date: '2021-08-24'
-modification_date: '2026-05-13'
+modification_date: '2026-08-30'
 author: Mauricio Velazco, Splunk
 status: production
 type: Hunting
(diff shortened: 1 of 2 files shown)


nasbench (approved):
LGTM. Thanks for the quick fix

PR #4247, opened 2026-08-31
Sigma-to-SPL Translation: macOS osascript Detection

rviskova13:
### Details
This detection idea follows a research focused on the abuse of osascript on macOS. It is equivalent of cmd.exe on macOS, allowing to execute various commands, code pieces and scripts. And it is often abused by attackers.

This particular detection focuses on detecting osascript using "display dialog/alert" method, attempting to display fake system dialogs or credential prompts. It is a common social engineering tactic on macOS - attackers leverage AppleScript for credential harvesting or malware delivery.

### DIFF
--- a/detections/endpoint/macos_utility_osascript_prompting_for_user_interaction.yml
+++ b/detections/endpoint/macos_utility_osascript_prompting_for_user_interaction.yml
@@ -2,7 +2,7 @@ name: MacOS Utility Osascript Prompting for User Interaction
 id: d565ae18-7db6-460c-a2eb-a7109f18bc5e
 version: 1
 creation_date: '2026-08-03'
-modification_date: '2026-08-28'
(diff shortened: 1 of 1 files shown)


detections/endpoint/macos_utility_osascript_prompting_for_user_interaction.yml
```diff
@@ -0,0 +1,88 @@
+name: MacOS Utility Osascript Prompting for User Interaction
+id: d565ae18-7db6-460c-a2eb-a7109f18bc5e
+version: 1
+creation_date: '2026-08-03'
+modification_date: '2026-08-28'
+author: Radka Viskova, Splunk
+status: experimental
```
nasbench: All detections with a valid test must be production
```suggestion
status: production
```

PR #4261, opened 2026-09-07
Detection Against Content Copied From Browsers Executed through Run

onurmerdogan:
### Details

This detection correlates RunMRU registry modification and clipboard change from browsers. Correlation is done through the hash of the content that was copied to clipboard.

These activities combined, gives a solid base for detecting ClickFix attacks. Adversaries either asks users to manually copy or use Javascript to modify the clipboard by themselves. It first triggers the clipboard change event by the browser process. Once victim executes the copied content through Run dialog box, it triggers the modification of the RunMRU registry key.

Sysmon calculates the content hash based on UTF-16 LE encoding. Therefore RunMRU value is trimmed of its suffix and converted to UTF-16 LE, before its hash was calculated. I've also added missing Sysmon EID 24 to the repository.

### DIFF
--- a/data_sources/sysmon_eventid_24.yml
+++ b/data_sources/sysmon_eventid_24.yml
@@ -68,28 +68,17 @@ fields:
     - tag
     - tag::eventtype
 output_fields:
-    - action
     - dest
     - dvc
-    - dvc_nt_host
     - event_id
-    - eventtype
-    - linecount
-    - os
     - process_exec
     - process_guid
     - process_hash
     - process_id
     - process_name
     - process_path
-    - punct
     - severity
-    - severity_id
-    - signature
-    - signature_id
     - splunk_server
-    - src_host
     - user
-    - user_id
     - vendor_product
-example_log: <Event xmlns='http://schemas.microsoft.com/win/2004/08/events/event'><System><Provider Name='Microsoft-Windows-Sysmon' Guid='{5770385f-c22a-43e0-bf4c-06f5698ffbd9}'/><EventID>24</EventID><Version>5</Version><Level>4</Level><Task>24</Task><Opcode>0</Opcode><Keywords>0x8000000000000000</Keywords><TimeCreated SystemTime='2026-09-07T14:37:29.0002181Z'/><EventRecordID>274447</EventRecordID><Correlation/><Execution ProcessID='3400' ThreadID='4440'/><Channel>Microsoft-Windows-Sysmon/Operational</Channel><Computer>EC2AMAZ-FKEOM7G</Computer><Security UserID='S-1-5-18'/></System><EventData><Data Name='RuleName'>-</Data><Data Name='UtcTime'>2026-09-07 14:37:28.999</Data><Data Name='ProcessGuid'>{1b43529b-bf4f-6a9e-e4eb-020000006600}</Data><Data Name='ProcessId'>8988</Data><Data Name='Image'>C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe</Data><Data Name='Session'>2</Data><Data Name='ClientInfo'>user: EC2AMAZ-FKEOM7G\Administrator hostname: OERDOGAN-M-7N5Y</Data><Data Name='Hashes'>MD5=F2CD63BD7E2B5F9B738F589AF3BB1F8F,SHA256=397EE4CA855D96072D57E89ABFCA9F83634DAEA133E5302CB0766E87FDC3FEFC,IMPHASH=00000000000000000000000000000000</Data><Data Name='Archived'>false - failed with error 1337</Data><Data Name='User'>EC2AMAZ-FKEOM7G\Administrator</Data></EventData></Event>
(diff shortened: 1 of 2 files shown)


data_sources/sysmon_eventid_24.yml
```diff
@@ -0,0 +1,95 @@
...
+    - punct
+    - severity
+    - severity_id
+    - signature
+    - signature_id
+    - splunk_server
+    - src_host
+    - user
+    - user_id
+    - vendor_product
```
nasbench, line 83: For this select the fields that *must be in the output*. Not everything in here should be imo.

detections/endpoint/windows_content_copied_from_browser_was_executed.yml
```diff
@@ -0,0 +1,96 @@
...
+        "sidekick.exe", "thorium.exe", "torch.exe", "vivaldi.exe", "wavebrowser.exe", "whale.exe", "zen.exe"))
+    | `security_content_ctime(firstTime)`
+    | `security_content_ctime(lastTime)`
+    | `windows_content_copied_from_browser_was_executed_filter`'
+how_to_implement: |-
+    This detection requires Sysmon event logs. Configure your environment to ingest Sysmon registry modification (Event ID 13)
+    and clipboard change (Event ID 24) events. Ensure that registry modification events are collected for the RunMRU registry key
+    and clipboard change events are collected from common browser processes. Ingest the data through the appropriate Splunk technology
+    add-on and normalize field names using the Splunk Common Information Model (CIM).
+known_false_positives: No false positives were known for this detection.
```
nasbench: We are using a template for this phrase.

```suggestion
known_false_positives: No false positives have been identified at this time.
```

detections/endpoint/windows_content_copied_from_browser_was_executed.yml
```diff
@@ -0,0 +1,96 @@
...
+intermediate_findings:
+    entities:
+        - field: dest
+          type: system
+          score: 40
+          message: Content $process$ was copied from $process_name$ and was executed through the Run dialog box on $dest$
+        - field: user
+          type: user
+          score: 40
+          message: Content $process$ was copied from $process_name$ and was executed through the Run dialog box by $user$
```
nasbench: One should be enough as per the convention that was set during 6.0 release. Except if you think that there should 2 risk object emitted for both the machine/user. Lmk

```suggestion
```

detections/endpoint/windows_content_copied_from_browser_was_executed.yml
```diff
@@ -0,0 +1,96 @@
...
+      latest_offset: $info_max_time$
+    - name: View risk events for the last 7 days for - "$dest$"
+      search: '| from datamodel Risk.All_Risk | search normalized_risk_object IN ("$dest$") | stats count min(_time) as firstTime max(_time) as lastTime values(search_name) as "Search Name" values(risk_message) as "Risk Message" values(analyticstories) as "Analytic Stories" values(annotations._all) as "Annotations" values(annotations.mitre_attack.mitre_tactic) as "ATT&CK Tactics" by normalized_risk_object | `security_content_ctime(firstTime)` | `security_content_ctime(lastTime)`'
+      earliest_offset: 7d
+      latest_offset: "0"
+intermediate_findings:
+    entities:
+        - field: dest
+          type: system
+          score: 40
```
nasbench: There is a convention for initial scoring. TTPs get 50 and anomalies get 20. See https://github.com/splunk/security_content/wiki/6-%E2%80%90-How-are-risk-score-calculated-for-Event-Based-Detections
```suggestion
          score: 50
```

detections/endpoint/windows_content_copied_from_browser_was_executed.yml
```diff
@@ -0,0 +1,96 @@
...
+          message: Content $process$ was copied from $process_name$ and was executed through the Run dialog box on $dest$
```
nasbench:
```suggestion
          message: The content [$process$] was copied from [$process_name$] to the clipboard and was executed through the Run dialog box on [$dest$]
```

detections/endpoint/windows_content_copied_from_browser_was_executed.yml
```diff
@@ -0,0 +1,96 @@
...
+product:
+    - Splunk Enterprise
+    - Splunk Enterprise Security
+    - Splunk Cloud
+category: endpoint
+security_domain: endpoint
+tests:
+    - name: True Positive Test
+      attack_data:
+        - data: https://media.githubusercontent.com/media/splunk/attack_data/refs/heads/master/datasets/attack_techniques/T1202/windows_browser_copy_paste/windows_browser_copy_paste.log
```
nasbench:
```suggestion
        - data: https://media.githubusercontent.com/media/splunk/attack_data/master/datasets/attack_techniques/T1202/windows_browser_copy_paste/windows_browser_copy_paste.log
```

nasbench (approved):

PR #4263, opened 2026-09-08
Tune and Add More Context to Noisy Detections

nasbench:
This PR updates 2 analytics based on telemetry results.

### Updated Analytics

-  `Detect Excessive User Account Lockouts` - Update logic slightly and enhance metadata sections to make it clear on how this should be used and what type of FPs are expected.
- `Prohibited Network Traffic Allowed` - Enhance metadata to indicate how the analytic works and how it should be used.

### New Data Source

- `Windows Event Log Security 4740`

### DIFF
--- a/data_sources/windows_event_log_security_4740.yml
+++ b/data_sources/windows_event_log_security_4740.yml
@@ -1,0 +1,97 @@
+name: Windows Event Log Security 4740
+id: 387936fa-efa9-425d-9f1b-487425250ea9
+version: 1
+creation_date: '2026-09-08'
+modification_date: '2026-09-08'
+author: Nasreddine Bencherchali, Splunk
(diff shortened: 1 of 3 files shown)


onurmerdogan (approved):

nasbench:
After talking with Eric, the failure is unknown at this point. I will be merging and testing this internally. If things truly breaks. Will revert it.

pyth0n1c:
> After talking with Eric, the failure is unknown at this point. I will be merging and testing this internally. If things truly breaks. Will revert it.

Chiming in to confirm that it's unclear why this test is failing for us. It appears to work in multiple other test environments and, because we cannot log directly into GitHub runners, we cannot perform additional diagnosis on why this is failing in Github. It may be a performance or ingest-related issue that could be resolved by a faster runner or more runtime.
We will evaluate this PR after merge, as Nas noted above, in our other testing systems.

PR #4266, opened 2026-09-11
Automated Data Source TA and install.yml App Update 758

pyth0n1c:
This PR contains data source supported_TA and install.yml app version updates generated by contentctl-ng.

The workflow runs:

```bash
contentctl-ng update-data-source-tas --content-dirs .
```

PR #4268, opened 2026-09-12
Automated Data Source TA and install.yml App Update 759

pyth0n1c:
This PR contains data source supported_TA and install.yml app version updates generated by contentctl-ng.

The workflow runs:

```bash
contentctl-ng update-data-source-tas --content-dirs .
```


nasbench:
See for reason for fail https://github.com/splunk/security_content/pull/4266#pullrequestreview-5183396522

Will keep this open as it has more updates.

PR #4276, opened 2026-09-21
Update EventID 10 Analytics

nasbench:
This PR updates analytics leveraging the EID 10 data source by changing the logic on how Access Masks are compared. Instead of direct string matching which is limited and could be bypassed, we use bitwise operators to look for specific masks inside the granted access mask. This way we can get an accurate picture of whats being requested.

This will of course results in more FPs now that the analytics are technically more accurate. For this, I did not do a logic/FP analysis this time around but more of simply fixing the logic. In the next releases I will be investigating matches and evaluating the logic/accuracy and if the content is still needed.


detections/endpoint/detect_credential_dumping_through_lsass_access.yml
```diff
@@ -38,11 +59,14 @@ drilldown_searches:
       earliest_offset: 7d
       latest_offset: "0"
 finding:
-    title: The $SourceImage$ has attempted access to read $TargetImage$ was identified on endpoint $dest$, this is indicative of credential dumping and should be investigated.
+    title: The process [$SourceImage$] requested read access via the access mask [$GrantedAccess$] to [$TargetImage$] on [$dest$], which may indicate credential dumping.
```
onurmerdogan: `GrantedAccess` will not render, since it is not part of the final table, I think you should replace it with `granted_access`, this persist in all the other detections also.

nasbench: Nice catch. Updated all other analytics as well.

onurmerdogan (approved):

